PR #6, opened 2026-05-10
fix(tui): clean up terminal on quit so mouse tracking doesn't leak to host shell

PengyuChen01:
## Summary

Follow-up to #4. After quitting kobe via Ctrl+C×2 (or `q`), the host shell receives a non-stop stream of `\x1b[<...M` SGR mouse events from every cursor move, and the alt-screen is left unrestored. Screenshot from this morning's bug report showed the host prompt buried under TUI residue plus a wall of mouse coordinate triples.

## Root cause

`useKobeKeybindings`'s default `onQuit` was `() => process.exit(0)`. `process.exit()` is a synchronous Node-level kill — it bypasses `beforeExit`, signal listeners, and every opentui exit hook. Without invoking `renderer.destroy()` first, the renderer never writes the terminal-restore sequences (`\x1b[?1003l`, `\x1b[?1006l`, `\x1b[?1049l`, `setRawMode(false)`) before stdout closes, so the TTY stays in raw + mouse-tracking mode.

Pre-#4 this never surfaced for Ctrl+C because opentui's built-in `exitOnCtrlC: true` handler called `destroy()` itself. Once #4 set `exitOnCtrlC: false` and routed Ctrl+C through `onQuit`, the unguarded `process.exit(0)` got hit. The `q` quit path has had this bug since day one (same default), but the confirm dialog masked it visually and most users wouldn't trigger it during heavy mouse activity.

## Fix

```diff
-  const onQuit = opts.onQuit ?? (() => process.exit(0))
+  const onQuit =
+    opts.onQuit ??
+    (() => {
+      renderer?.destroy()
+      process.exit(0)
+    })
```

`renderer.destroy()` walks `cleanupBeforeDestroy` → native `destroyRenderer`, which emits the disable sequences through stdout synchronously. They're flushed before `process.exit(0)` ends the loop, so the shell gets a clean TTY when control returns. Both Ctrl+C×2 and `q`-confirm paths fixed in one change since they share the default.

## Test

Adds a regression test to `test/behavior/ctrl-c-quit.test.ts`:

- Drives Ctrl+C×2, waits for pty close
- Asserts the raw byte buffer contains a mouse-disable sequence (any of `\x1b[?100Nl`) — would have NONE without `destroy()`
- Asserts alt-screen-exit (`\x1b[?1049l`) is present

Verified locally:
- [x] Old code (without `renderer.destroy()`) → new test fails (no mouse-disable in buffer)
- [x] New code → all 3 ctrl-c-quit tests pass
- [x] `bun run typecheck` clean
- [x] `bunx biome check` clean on touched files

## Summary by CodeRabbit

* **Bug Fixes**
  * Quitting the application (double Ctrl+C or 'q') now properly restores the terminal state, including disabling mouse tracking and restoring normal display mode.

[Review Change Stack](https://app.coderabbit.ai/change-stack/Sma1lboy/kobe/pull/6)

### DIFF
--- a/packages/kobe/CHANGELOG.md
+++ b/packages/kobe/CHANGELOG.md
@@ -26,6 +26,12 @@ and the project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.
   exit" hint in the status bar); a second Ctrl+C within 1.5s exits.
   Matches the standard TUI muscle memory used by claude-code, fish,
   and ipython.
+- Quitting kobe (via Ctrl+C×2 or `q`) now restores the host terminal
+  cleanly — previously `process.exit(0)` skipped opentui's teardown,
+  leaving mouse tracking enabled (the host shell would receive a
+  stream of `\x1b[<…M` SGR mouse events from every cursor move) and
+  the alt-screen unrestored. The default `onQuit` now calls
+  `renderer.destroy()` first.
 
 ## [0.1.1] - 2026-05-09
 
--- a/packages/kobe/src/tui/context/keybindings.ts
+++ b/packages/kobe/src/tui/context/keybindings.ts
@@ -202,7 +202,15 @@ export function useKobeKeybindings(opts: KobeKeybindingsOpts): void {
   const onQuit =
     opts.onQuit ??
     (() => {
-      renderer?.destroy()
+      // Defense in depth: if destroy() throws (FFI to native renderer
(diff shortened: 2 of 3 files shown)


PengyuChen01:
Addressed in 6a23ae1 — wrapped `renderer.destroy()` in try/catch, log the failure to stderr (safe under raw mode, visible in the next shell prompt), and force `process.exit(0)` unconditionally. The 3 existing behavior tests still pass.

PR #9, opened 2026-05-10
fix(tui): surface ctrl+j newline fallback in composer hint

PengyuChen01:
## Summary

- **Issue:** pressing Shift+Enter in the composer submits the prompt instead of inserting a newline on most terminals (Windows Terminal, default Terminal.app, default gnome-terminal, default tmux). The byte coming in is indistinguishable from plain Enter, so the `shift+enter → newline` binding is dead and the `return → submit` binding wins.
- **Why:** `shift: true` only appears on the key event when the terminal honors either kitty-keyboard / CSI-u or xterm modifyOtherKeys. opentui already opts into both, but many common terminals support neither. At the byte level we cannot recover the lost modifier, so there is no software-only way to make Shift+Enter work universally.
- **Fix:** advertise the universal fallback (Ctrl+J) directly in the composer's action hint, and tighten the keybindings docstring so the terminal-side limitation is clear for the next reader. Behavior on capable terminals is unchanged.

## Changes

- `Composer.tsx`: action hint now reads `enter send · shift+enter or ctrl+j newline · shift+tab mode`.
- `composer/keybindings.ts`: docstring expanded to call out modifyOtherKeys parsing, the dead-binding case, and that Ctrl+J is the workaround.

## Test plan

- [x] `bun --filter @sma1lboy/kobe typecheck` — clean.
- [x] `bunx biome check` on changed files — clean.
- [ ] Manual: launch `bun run dev` in a non-CSI-u terminal, confirm hint shows `ctrl+j` and Ctrl+J inserts a newline (existing `linefeed` binding, untouched).
- [ ] Manual: launch in kitty / wezterm / recent Windows Terminal, confirm Shift+Enter still inserts a newline.

### DIFF
--- a/packages/kobe/src/tui/panes/chat/Composer.tsx
+++ b/packages/kobe/src/tui/panes/chat/Composer.tsx
@@ -558,7 +558,7 @@ export function Composer(props: ComposerProps) {
   const actionHint = () => {
     if (!props.hasTask) return ""
     if (props.isStreaming) return "streaming — wait for done"
-    return "enter send · shift+enter newline · shift+tab mode"
+    return "enter send · shift+enter or ctrl+j newline · shift+tab mode"
   }
   const modelLabel = () => props.modelLabel?.() ?? "claude-code"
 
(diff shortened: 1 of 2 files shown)


PengyuChen01:
Reopening after deeper investigation — initial hypothesis was wrong. Claude Code's TUI handles Shift+Enter natively in the same terminal, so this is a kobe/opentui setup issue, not a hard terminal-side limitation. Closing while I dig into the actual root cause.

Issue #179, opened 2026-06-29
Exit the terminal layout error


bluetoblack9:
**Describe the bug**
When I execute exit on both terminal interfaces of a task, I find that there is a layout disorder. Clicking on other task will reset the layout. But when it comes to the last edited task, the layout is still incorrect and needs to be dragged again for window layout.

**To Reproduce**
1. new Task
2. middle terminal exit
3. right terminal exit
4. look terminal display

**Expected behavior**
Keep layout proportion or switch to default task

**Screenshots**

**Desktop (please complete the following information):**
 - OS:mac os 26.5
 - Browser no
 - Version mac os 26.5

**Additional context**

ZHallen122:
Hi @bluetoblack9,

Thank you for reporting this! We have confirmed the bug and my co-maintainer has already submitted a fix for it. 

You can track the progress in the PR here: #182 

We'll merge it soon and keep you updated. Thanks again for helping us improve!

Sma1lboy:
Hi @bluetoblack9 — confirmed and fixed. I verified this is a real bug; the fix is merged to main (#186). The layout now re-pins itself when a terminal pane exits, and while digging in I also fixed a couple of deeper issues in the same area (a global layout-state corruption triggered on shell exit, and the engine pane closing too eagerly on startup). We'll keep polishing the stability of this part of the code. Thanks a lot for the detailed report and the feedback!

PR #182, opened 2026-06-29
fix: heal a task's workspace layout when a pane exits (#179)

Sma1lboy:
## Direction

Bug / correctness — fixes open issue #179 ("Exit the terminal layout error"). This was the only open issue not reserved as `good first issue`.

## Problem

Reported in #179: after exiting **both** terminal interfaces of a task (typing `exit` in the middle + right workspace-split terminals), the layout becomes disordered, and the **currently-focused** task stays broken until you drag the panes back by hand. The reporter noted that *switching to another task resets the layout* — but the last-edited task never gets that switch, so it stays wrong.

Root cause: closing a pane destroys it and tmux hands its cells to a neighbouring pane, reflowing the absolute-width Tasks rail and the right column off their pinned geometry — the **same** disorder a terminal resize causes. But `window-resized` never fires on a pane close (the window size is unchanged), and the only hook that does fire (`window-layout-changed`) is wired to *capture* a drag, not to *heal*. So nothing re-pins the focused task's layout. Switching tasks worked only because switch-in runs `healWorkspaceLayout` directly.

## Fix

Wire tmux's `pane-exited` hook to the existing `heal-layout` handler (`set-hook -g pane-exited`), reusing the exact idempotent, coalesced re-pin that a task switch already runs — the operation the reporter confirms fixes the disorder. It re-pins the Tasks rail width + right-column geometry to the shared globals the moment a pane closes, and is a no-op for role-less sessions and for panes already at the target geometry. One added hook registration line plus comment/doc updates on the reused `heal-layout` handler; no behavior change to the existing resize/capture hooks.

## Verification

- `bun run lint` ✓ · `bun run typecheck` ✓ (both packages).
- `bun test` for the touched area — `terminal-tmux`, `layout-coord`, `terminal-pane-heal` — 64 pass / 0 fail.
- Empirically validated against a live tmux 3.4 server that the `pane-exited` hook fires on a pane `exit`. Confirmed via the upstream tmux CHANGES that `pane-exited` has existed since tmux **2.2**, well below kobe's documented ≥3.2 minimum, so adding it to the session-setup batch (which halts on the first failed command) is safe on every supported tmux.

## Follow-ups / deferred

- I could not reproduce the exact *visual* disorder end-to-end in the cloud checkout (no live TUI), so this relies on the reporter's own observation that the existing `healWorkspaceLayout` — the operation now triggered on pane exit — resets the layout when invoked via a task switch. Worth a manual confirm on the reporter's macOS setup from #179.
- `pane-exited` also fires on chat-tab close / task delete; the heal is idempotent and coalesced so those are harmless no-ops, but if pane-churn-heavy flows ever show overhead, the coalesce window in `layout-coord.ts` is the knob.

---
_Generated by [Claude Code](https://claude.ai/code/session_01UPsPUNg9y4aL7TdjxBm9Lk)_

### DIFF
--- a/.changeset/heal-layout-on-pane-exit.md
+++ b/.changeset/heal-layout-on-pane-exit.md
@@ -1,0 +1,5 @@
+---
+"@sma1lboy/kobe": patch
+---
+
+fix: a task's workspace layout now re-pins itself after you exit a terminal pane. Closing a workspace-split terminal (typing `exit`) hands its cells to a neighbouring pane, which knocks the fixed-width Tasks rail and the right column off their pinned geometry — the same disorder a terminal resize causes, except `window-resized` never fires because the window size is unchanged. Until now the only recovery was switching to another task and back (which heals on switch-in), so the currently-focused task stayed visually broken until you dragged the panes back yourself. kobe now heals the layout on tmux's `pane-exited` hook, re-pinning the rail and right column to the shared globals the moment a pane closes.
--- a/packages/kobe/src/cli/index.ts
+++ b/packages/kobe/src/cli/index.ts
@@ -612,11 +612,14 @@ async function main(): Promise<void> {
     return
   }
   if (subcommand === "heal-layout") {
-    // `window-resized` tmux hook handler: re-pin the resized session's Tasks-rail
-    // width + right-column geometry to the shared globals. Fixes the first-attach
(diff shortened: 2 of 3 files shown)


Sma1lboy:

PR #262, opened 2026-07-06
fix: keep the embedded terminal cursor-row floor so the block cursor lands in its true column

Sma1lboy:
## Direction

Bugs / correctness — found by direct code review of the LIVE embedded-terminal pane (issue #16), specifically the previously **untested** row-snapshot builder `src/tui/panes/terminal/xterm-chunks.ts`.

## Problem

`refreshSnapshot()` in `pty.ts` renders each PTY row with `xtermLineToChunks(line, minLast)`, passing `minLast = active.cursorX - 1` on the cursor's row. `minLast` is a **floor**: the cursor row must materialize out to at least the cursor column so the block-cursor overlay (`overlayCursorRow`) has real cells to land on — `overlayCursorRow` only inverts a cell if the cursor column falls *inside* the row's materialized width; otherwise it appends a single inverse space at the **end** of the text.

The visible-cell scan discarded that floor:

```ts
let last = Math.min(line.length - 1, minLast)   // floor
for (let x = 0; x < line.length; x++) {
  ...
  if (isVisibleCell(cell)) last = x              // plain assignment — floor lost
}
```

Because `x` is monotonically increasing, the assignment correctly finds the last visible index but **overwrites** the `minLast` floor whenever the row's content ends before the cursor. Result: the row collapses to the last glyph, and the cursor overlay snaps left onto that glyph instead of rendering at its true column.

**Concrete trigger:** row `"hello"` in cols 0–4, cursor parked at col 20 (`minLast = 19`). Buggy output materializes only `"hello"` (5 cols); the cursor renders at col 5 instead of col 20 — off by 15. This shows up on trailing-whitespace-before-cursor, cleared-to-end-of-line prompts, right-aligned status lines, and completion menus.

## Fix

Take the max so the floor survives the scan:

```ts
if (isVisibleCell(cell)) last = Math.max(last, x)
```

One line, one function. Non-cursor rows (`minLast = -1`) are unchanged: `Math.max(-1, x)` is identical to the old assignment.

## Verification

- Added `test/tui/terminal-xterm-chunks.test.ts` (the module had no test): pins the floor behavior, the "no floor" baseline, all-blank collapse, floor-never-shrinks, and floor-clamped-to-width. Confirmed the floor + clamp tests **fail** against the pre-fix code and pass with the fix.
- `bun run lint` — green (660 files).
- `bun run typecheck` — green (`@sma1lboy/kobe-daemon` + `@sma1lboy/kobe`).
- `bun test` terminal render/viewport suites — 58 pass.

## Follow-ups

- None required. The sibling render path (`overlayCursorRow`) is correct as-is; the defect was solely the lost floor in the scan.

---
_Generated by [Claude Code](https://claude.ai/code/session_01Ho4DV4xKibiRRyZPP21jrv)_


Sma1lboy:
The one-line fix (last = Math.max(last, x)) already landed on main (xterm-chunks.ts:169,226); this branch is 300+ files stale. Its unit tests pinning the cursor-row floor were ported to main directly (c2ced0b5).

PR #268, opened 2026-07-07
feat: CJK-aware word boundaries in TUI text inputs

Sma1lboy:
## Summary
- opentui's native (Zig) word boundaries are useless for Chinese: a CJK run is walked in arbitrary fixed-width hops, and punctuation (`]`, `=`, `】`…) never splits. kobe is Simplified-Chinese-default, so alt+←/→ and alt+backspace inside a Chinese prompt are a core editing path.
- Patch `TextareaRenderable.prototype` once at host boot (`installCjkWordBoundaries`, wired in `host-boot.tsx`): same-line word move/delete are reimplemented via `Intl.Segmenter("zh", { granularity: "word" })` — 「中文输入法优化」 hops 中文|输入|法|优|化, and ASCII/fullwidth punctuation are natural split points (a punctuation run is one hop, vim-style). One patch point covers every `<input>`/`<textarea>` (quick-task composer, new-task dialog, rename dialog, settings) without touching the >500-line dialog files.
- Coordinate contract probed against @opentui/core 0.4.3: edit-buffer cursor col/offset are terminal CELLS (CJK glyph = 2, mid-glyph snaps down). The pure column math lives in `src/tui/lib/cjk-word.ts`; `displayWidth`/`charWidth` moved from `cli/export-cmd.ts` to `src/lib/display-width.ts` so the TUI doesn't import from the CLI layer (re-exported, existing imports keep working).
- At a line edge (multi-line `<textarea>` cross-line moves) the original native method still runs; selection bookkeeping mirrors the originals.

## Test plan
- [x] `test/tui/cjk-word.test.ts` — 16 unit tests on the pure boundary math (CJK cell widths, mid-glyph snap, `]`/`=`/fullwidth splits, punctuation runs, whitespace skip, mixed CJK+ASCII prompt round-trip)
- [x] Integration smoke on a real `TextareaRenderable` via `@opentui/core/testing`: backward hops land per Chinese word (`end → ] → 值 → = → 化 → 优 → 法 → 输入 → 中文`); repeated alt+backspace deletes one unit at a time
- [x] Full kobe suite (218 tests), root `bun run lint`, package typecheck — all green

coverage-exemption: packages/kobe/src/tui/lib/cjk-word-patch.ts — imports @opentui (not loadable under vitest); pure boundary math is tested in test/tui/cjk-word.test.ts

### DIFF
--- a/.changeset/cjk-word-boundaries.md
+++ b/.changeset/cjk-word-boundaries.md
@@ -1,0 +1,5 @@
+---
+"@sma1lboy/kobe": patch
+---
+
+TUI 输入框中文分词:alt+←/→ 跳词、alt+backspace 删词现在按中文词语(Intl.Segmenter)移动,`]`、`=`、全角标点都是词边界;之前 native 词边界对中文是固定步长乱跳。覆盖所有 `<input>`/`<textarea>`(quick-task、new-task、rename、settings)。
--- a/packages/kobe/src/cli/export-cmd.ts
+++ b/packages/kobe/src/cli/export-cmd.ts
@@ -98,62 +98,11 @@ function renderCsv(rows: readonly Record<Column, string>[]): string {
   return lines.join("\n")
 }
 
-/**
- * Terminal display width of `s` in cells. Iterates by code point (so astral
- * characters — CJK Extension B, emoji — count once, not as two UTF-16 units),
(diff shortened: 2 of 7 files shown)


Sma1lboy:
本地迭代阶段,先不走 PR 流程;分支保留,后续需要时可 reopen。

PR #270, opened 2026-07-07
feat: unified process-identity model for terminal tabs

Sma1lboy:
## What

Every tab is a shell; an engine (claude/codex) is just a process running in it. One state model, not two component kinds.

- **OSC 0/2 title tracking** on all three PTY backends (`onTitleChange`, xterm-headless native for the Bun backend, shared `extractOscTitle` for pipe/mock)
- **`turn-polls` owns process identity end to end**: live title → `vendorFromTerminalTitle` (engine-owned matching in the registry) → turn detector attach/detach. A user-typed `claude` inside a plain shell gets the same ●/✓ turn chip as a kobe-launched engine tab, and loses it when the process exits
- **Tab default names are `$process $ordinal`** ("claude 3", "codex 4", "shell 5", "vim 2"), never "tab N"; engine titles normalize to one vocabulary ("✳ Claude Code" → "claude") via `titleDisplayName`
- **Restart contract**: `rehydrateTabs` keeps every tab — engine tabs resume their conversation, command tabs (degraded shells, dead editors) respawn as shells

## Fixes

- A single tab degraded to a shell no longer reopens as a fresh claude after restart
- Closing the engine leaf inside a split no longer leaves a stale turn chip flapping against the released PTY
- The split corner name tag hides when only one leaf survives (the tab label already carries the same name)
- Corner tags no longer show raw decorated titles ("✳ Claude Code")

## Coverage exemptions

coverage-exemption: packages/kobe/src/tui/panes/terminal/pty.ts — Bun-PTY IO backend (Bun.spawn terminal); not mountable under vitest, ~1% coverage predates this change
coverage-exemption: packages/kobe/src/tui/panes/terminal/pty-pipe.ts — child-process pipe fallback backend, same IO-bound nature, ~1% coverage predates this change
coverage-exemption: packages/kobe/src/tui/workspace/turn-polls.ts — Solid component-body hook (createEffect/onCleanup) requiring an opentui mount, which needs bun not vitest; 0% coverage predates this change

### DIFF
--- a/.changeset/terminal-process-identity.md
+++ b/.changeset/terminal-process-identity.md
@@ -1,0 +1,5 @@
+---
+"@sma1lboy/kobe": patch
+---
+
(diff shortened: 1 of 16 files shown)


Sma1lboy:
Merged locally (fast local-branch workflow for this dev stream); branch content is on local main.

PR #271, opened 2026-07-07
fix: react sidebar/filetree parity — stop chip clicks bubbling to pane focus-grab

Sma1lboy:
Parity audit of the React sidebar + filetree ports against their Solid originals (drift window: since c29e9d82 / the initial port).

Audit findings: the ports already share all framework-free logic via `../../tui/**` imports, so most Solid-side fixes landed automatically. Exactly one behavioral gap existed on each pane, the same class of bug: chip `onMouseUp` handlers (file tree header's Zen / Create-PR chips, sidebar panel's ZEN click) were missing `stopPropagation()`, so the click bubbled into the host pane's focus-grab `onMouseUp` and could instantly cancel the action — the same fix the Solid side already carries.

Everything else (Sidebar.tsx, row-cards, hover-tooltip, keys, FileTree rows/pane-core) verified line-by-line at parity; no changes needed.


Sma1lboy:
Merged locally (fast local-branch workflow for this dev stream); branch content is on local main.

Issue #274, opened 2026-07-08
kobe unavailable


bluetoblack9:
**Describe the bug**
During the use of claude, the task list suddenly disappears and cannot be restored

**To Reproduce**
no

**Expected behavior**
Do not crash the task。

**Screenshots**

<img width="957" height="595" alt="Image" src="https://github.com/user-attachments/assets/f6047d76-40e8-408c-8058-8c58adf78ed9" />

**Environment**
 - OS: macOS 26.5
 - Terminal app: Terminal.app
 - Where it happens: both  [main kobe TUI / inside a task (tmux) /。both]

**`kobe doctor` output (required)**
Run `kobe doctor` in the same terminal where the bug happens and paste the full output — it includes your kobe version, terminal, and keyboard-protocol info we need for triage.
See document

```text
paste `kobe doctor` output here
```

**Additional context**
List of tasks that are not displayed in the delivery directory. Create a new task and cannot use Kobe

I have tried the following operations, but still cannot function properly
1.kobe reset
2.kill kobe
3.kill bun
4.Restart Mac
5.new task

[kobe.log](https://github.com/user-attachments/files/29793624/kobe.log)
@Sma1lboy

bluetoblack9:
Is there any use of communication groups for communication

bluetoblack9:
I deleted the codeKo task from the picture.
The task list now shows the tasks that crashed before.
UI display issue?

PR #276, opened 2026-07-08
fix: back up a corrupt tasks.json before recovering to an empty index

Sma1lboy:
## Direction

Bugs / correctness — found by direct code review of the task-manifest persistence layer (`orchestrator/index/store.ts`), the store behind the TUI task list. Adjacent to the class of failure reported in #274 ("the task list suddenly disappears and cannot be restored").

## Problem

When `tasks.json` is unparseable — the shape a crash or half-write produces — `TaskIndexStore.load()` recovers by booting with an **empty** index and returns to the UI immediately. Its comment claims *"The stale file is left in place,"* implying the data is still there to recover.

It isn't, for long. `readDiskTasks()` (the read-merge-write base used by every `save()`) also treats the corrupt file as `[]`. So the **next `save()`** — any task create, status flip, reorder, PR-status refresh — rewrites `tasks.json` from that empty base via `mergeWithDisk`, **permanently destroying** every task row and the worktrees/branches those rows pointed at. Once that write lands, the data is gone with no way back.

## Fix

Before recovering to empty, copy the raw bytes to a timestamped sibling `tasks.json.corrupt-<ISO>` via a new `backupCorruptManifest` helper. The startup warning now names that backup so a user (or `/file-issue`) can recover their task list by hand. The backup is best-effort: a failed write logs and returns `null` rather than throwing, preserving the "a corrupt manifest must never wedge boot" policy.

## Refactor (forced by the file-size cap)

`store.ts` was already 672 lines — over the ~500 cap. Touching it means owning the shrink, so this PR extracts the **pure, stateless** load-side half (version constant, `normalizeIndex`, `coerceTask`, `coercePRStatus`, the PR/status type guards, and the new `backupCorruptManifest`) into a sibling `orchestrator/index/normalize.ts`. The functions moved verbatim — no behavior change — leaving `store.ts` at exactly 500 lines and focused on the read-merge-write lifecycle. A dead `EMPTY_INDEX` const (`void EMPTY_INDEX`) was dropped in the process.

## Verification

- Added two tests to `test/orchestrator/store-load-edge.test.ts`: corrupt JSON now writes a `tasks.json.corrupt-*` backup holding the original bytes; and the end-to-end "corrupt → recover → a create clobbers tasks.json → data still recoverable from the backup" scenario. Confirmed the backup assertion **fails** against the pre-fix code.
- `bun run lint` — green (668 files).
- `bun run typecheck` — green (`@sma1lboy/kobe-daemon` + `@sma1lboy/kobe`).
- `bun test test/orchestrator/` store suites — 41 pass across the load-edge / concurrency / heal / remove / reorder files (the 3 unrelated `active-task.test.ts` failures are a pre-existing `vi.stubEnv` bun-vs-vitest runner gap, present on `main`).
- File-size cap: `store.ts` 500 lines, `normalize.ts` 187 lines — both under the gate.

## Follow-ups (deliberately deferred)

- The non-object-root / unsupported-version recovery paths in `normalizeIndex` also discard data, but they require syntactically-valid-yet-structurally-wrong JSON (which a crash won't produce — a crash truncates/garbles → parse failure), so they're out of scope for this correctness fix. Extending the backup to those paths would need `normalizeIndex` to surface a "recovered-to-empty vs legitimately-empty" signal, since it can't currently tell a garbage file from a valid empty index.
- The `readDiskTasks()` mid-session corrupt path (a peer corrupts the file after a clean load) still silently reads `[]`; the load-time backup already covers the common startup case.

---
_Generated by [Claude Code](https://claude.ai/code/session_01GefRHHyt57hCGg2B9tefjq)_


Sma1lboy:
Ported to main directly (c2ced0b5): the branch conflicted structurally with the store-codec.ts split that landed later — the essential fix (back up corrupt tasks.json bytes before empty-index recovery, on both the load and save-time read paths) is now applied against the current structure, with the corrupt-then-save regression test.

PR #278, opened 2026-07-09
fix: split-leaf ctrl+w/F2 gating and live leaf-1 titles

Sma1lboy:
## Summary

Two split-view bugs, both in the React migration layer:

**ctrl+w (and F2) never reached a split leaf.** `TerminalTabs` (parent) and `TerminalSplit` (child) share the chords, resolved by the keymap's LIFO stack — but React registers bindings in mount effects (children first), so ancestors land on top, inverting the Solid-era precedence. `chat.tab.close` always won: on a single tab ctrl+w just toasted "cannot close last tab"; on multiple tabs it closed the whole tab instead of the leaf. F2 renamed the tab instead of the leaf. Fix: the tab-level close/rename entries gate themselves off while the active tab is split (>1 leaf), so the chords fall through the stack to the leaf bindings. Documented the gating rule in `keymap.ts` / `keybindings-chat.ts`.

**A shell tab's leaf-1 corner tag froze on "zsh".** The live-title (OSC 0/2) subscription excluded leaf-1, so entering claude/vim from a shell tab never updated its split name. Leaf-1 now subscribes too; naming prefers `engineTitle || liveTitle || basename`, so engine tabs keep their conversation title.

## Test plan
- `splitLeafNames` unit tests extended: leaf-1 live title wins without an engine title, engine title still outranks it
- typecheck, full test suite (226), root lint all green locally

coverage-exemption: packages/kobe/src/tui-react/lib/keymap.ts — comment-only change (header doc of the ancestor-on-top gating rule); no executable lines touched

### DIFF
--- a/.changeset/split-close-and-leaf-titles.md
+++ b/.changeset/split-close-and-leaf-titles.md
@@ -1,0 +1,7 @@
+---
+"@sma1lboy/kobe": patch
+---
+
+Fix ctrl+w/F2 not reaching split leaves, and split leaf-1's corner tag freezing on "zsh".
+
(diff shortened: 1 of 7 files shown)


Sma1lboy:
Merged locally (squash onto main as 1f8190a0) — solo-dev fast path, skipping the PR gate this round.

PR #280, opened 2026-07-09
fix: reassemble a PTY window title whose escape splits across a read boundary

Sma1lboy:
## Direction

Bugs / correctness — found by direct code review of the newest daemon code (commit `d5cbbd32`, "live-process tab titles + kobe api pty-list"), specifically the previously test-thin OSC-title scanner in `packages/kobe-daemon/src/daemon/pty-host.ts`.

## Problem

`scanTitle` tracks each session's last OSC 0/2 window title with a plain string scan (no VT emulation), carrying any unterminated tail into the next chunk so a title that straddles a PTY read boundary can be completed. The carry chose its start with:

```ts
const esc = rest.lastIndexOf("\x1b")
session.titleCarry = esc === -1 ? "" : rest.slice(esc, esc + TITLE_CARRY_CAP)
```

`lastIndexOf("\x1b")` returns the **last** raw ESC byte, which is wrong when the tail already contains a *second* ESC after the title's introducer. Two concrete splits trigger it (a PTY read boundary can fall on any byte):

- **Split ST terminator.** A title terminated by `ESC \` whose two bytes split across chunks leaves `rest = "\x1b]0;title\x1b"`. `lastIndexOf` picks the trailing terminator ESC, so the carry keeps only `"\x1b"` and **discards the whole `\x1b]0;title` introducer**. The next chunk (`"\\"`) then matches nothing and the title is lost.
- The same class of loss applies whenever the introducer's ESC is no longer the last ESC in the tail.

Symptom: the tab's live title (and the `pty.list` / `kobe api pty-list` inventory) silently fails to update on that title, until the next *complete* title arrives.

## Fix

Anchor the carry on the OSC **introducer** (`\x1b]`) — the byte the next chunk actually needs — instead of the last bare ESC, falling back to a lone trailing ESC (which may be the first byte of a split introducer). The scan/fold is extracted into a pure, exported `scanOscTitle(prevCarry, chunkText)` helper; the stateful UTF-8 `StringDecoder` stays in `PtyHost`. As a bonus the introducer-anchored carry no longer accumulates trailing color-escape junk between titles.

## Verification

- Added pure `scanOscTitle` unit tests to `test/render/pty-host.test.ts` (which already imports the daemon module): BEL + ST termination, last-title-wins, no-title→null, mid-body split, the **split-ST regression**, split-introducer, and introducer-anchored carry. The split-ST case fails against the pre-fix `lastIndexOf` carry and passes with the fix.
- `bun test test/render/pty-host.test.ts` — 13 pass (6 existing real-PTY tests + 7 new), so the full `onData → scanTitle` path is still exercised against a real Bun child.
- `bun run lint` — green (671 files). `bun run typecheck` — green (`@sma1lboy/kobe-daemon` + `@sma1lboy/kobe`).
- Note: `test/render/pty-hosted.test.ts` segfaults (a Bun 1.3.11 panic while mounting opentui in the headless sandbox) — reproduced identically on a clean `main`, so it is pre-existing and unrelated to this change.

## Follow-ups

- None required. A title interrupted *mid-body* by another escape sequence is unrecoverable by any string scan, but that interleaving doesn't occur in practice (OSC sequences are emitted atomically) — deliberately not handled.

---
_Generated by [Claude Code](https://claude.ai/code/session_014L92okfp425Kt1pjGtNz9a)_


Sma1lboy:
Ported to main directly (c2ced0b5): the original fix (b8737857) merged the same day this PR opened, then regressed in the #334 extraction to pty-observability.ts. The introducer-anchored carry is re-applied there as the pure foldOscTitle fold, with this PR's cross-chunk boundary tests.

PR #302, opened 2026-07-10
fix: chattab switch-back no longer loses claude's UI (park removal + reliable reattach repaint + replay reply mute)

Sma1lboy:
## The bug (as reported)

Switch away from a chattab → come back a while later → claude's input box / transcript not painted; the process is fine; typing a long enough message (forcing the input box to grow → full redraw) brings the content back. Intermittent.

## Root-cause chain (each link verified)

1. The park sweep releases a hidden tab's local xterm after 120s; switching back goes through the pty-host reattach path.
2. The reattach replay reconstructs the screen from the host's byte-capped ring buffer (512KB). In a long session, the frames that painted the static UI (claude's input box) predate the window — the replayed screen structurally lacks them.
3. The repaint wiggle that is supposed to fix exactly that (shrink one row, restore) fires its two resizes back-to-back. **Measured: two zero-gap resizes deliver ONE coalesced SIGWINCH and the child reads the already-restored (unchanged) size** — a size-comparing app like claude skips the repaint. Race → intermittent.
4. Bonus defect found on the same path: the fresh emulator answers the child's PAST terminal queries found in the replay (DSR/DA), injecting stray CPR bytes into claude's stdin (first commit).

## Fixes

- **No more auto-parking** (owner call): background tabs keep their local xterm resident; `parkIdle` stays for the perf harness. The everyday tab-switch never touches the replay path anymore.
- **Reliable wiggle** for the remaining replay path (TUI restart): after the shrink, wait for the child's repaint output (200ms timeout for children that don't repaint on WINCH) before restoring the size — two real SIGWINCHes with a real size change each.
- **Replay reply mute**: emulator auto-replies are suppressed for exactly the replay chunk's parse; live queries still answered.

## Tests

- Wiggle regression upgraded: the child now repaints only on an actual size change (claude's behavior) — it FAILS against the old zero-gap wiggle (verified red) and passes with the wait.
- Replay-reply regression: a replayed `ESC[6n` produces no reply; a live one after it produces exactly one.
- Full unit (243) + render (63) suites, typecheck, root lint green.

### DIFF
--- a/.changeset/full-identity-scrub.md
+++ b/.changeset/full-identity-scrub.md
@@ -1,0 +1,5 @@
+---
+"@sma1lboy/kobe": patch
+---
+
+Embedded terminals now scrub the outer emulator's entire identity namespace (LC_TERMINAL, ITERM_*, TERM_SESSION_ID, KITTY_*, GHOSTTY_*, WEZTERM_*, ALACRITTY_*, KONSOLE_*, VTE_VERSION, WT_*, TMUX, ZELLIJ, screen markers, __CFBundleIdentifier), not just TERM_PROGRAM. Apps with layered terminal detection (claude-code) no longer fall back to the outer emulator's dialect, which kept causing redraw artifacts inside kobe panes.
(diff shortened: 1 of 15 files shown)


Sma1lboy:
Landed locally on main by owner instruction (26c3a810 + 5bc74f8c cherry-picked; workflow: worktree → local merge, no PR).

PR #341, opened 2026-07-16
fix: keep external engine sessions out of Inbox

NarwhalChen:
## Summary

- keep cwd-matched external engine activity available to project status
- persist Inbox episodes only when an event has exact Kobe task and tab identity
- require tab identity at the AttentionInboxStore write boundary while retaining legacy task-level episode reads
- cover all four taskId/tabId truth-table cells with daemon handler regressions

## Verification

- bun run lint
- bun run typecheck
- bun run test: 2305 fast tests and 300 daemon tests passed
- bun run build
- 55 targeted Inbox/handler tests passed
- touched files remain below the 500-line cap

## Compatibility note

Legacy or manual hook events that provide taskId without KOBE_TAB_ID continue to update task activity but no longer create a task-level Inbox episode. Existing persisted tabId:null episodes remain readable and dismissible.

## Local environment note

The unrelated pure-tui-open-worktree behavior test intermittently misses its initial key because it waits for the PROJECTS paint rather than handler readiness. It passed in isolation and in PR CI; the final commit CI result is authoritative.

### DIFF
--- a/.changeset/calm-inbox-tabs.md
+++ b/.changeset/calm-inbox-tabs.md
@@ -1,0 +1,5 @@
+---
+"@sma1lboy/kobe": patch
+---
+
+Keep globally hooked Codex and Claude sessions outside Kobe out of the attention Inbox while preserving their cwd-matched project activity status.
--- a/.changeset/calm-inbox-tabs.md
+++ b/.changeset/calm-inbox-tabs.md
@@ -2,4 +2,4 @@
 "@sma1lboy/kobe": patch
 ---
 
-Keep globally hooked Codex and Claude sessions outside Kobe out of the attention Inbox while preserving their cwd-matched project activity status.
+Keep engine events without exact Kobe task-and-tab identity out of the attention Inbox while preserving their cwd-matched project activity status.
--- a/packages/kobe-daemon/src/daemon/attention-inbox.ts
+++ b/packages/kobe-daemon/src/daemon/attention-inbox.ts
@@ -105,6 +105,7 @@ export class AttentionInboxStore {
     detail: EngineActivityDetail | undefined,
     tabId: string,
   ): Promise<void> {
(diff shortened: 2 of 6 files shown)


NarwhalChen:
Addressed both Review CI Should-fix items in 1d68641b: added the missing no-taskId/with-tabId truth-table case, and tightened AttentionInboxStore.record so new writes require a tab identity while legacy task-level data remains readable.

NarwhalChen:
Followed up on the latest review in fc509e3b: clarified the activity-only fallback and the two-ID gate, enforced non-empty tab IDs at the Inbox store boundary, and added coverage for empty tab IDs plus the explicit-task AND tab truth-table case. I intentionally did not add logging for filtered global-hook events because those are expected and frequent outside Kobe tabs; logging them would add noise without indicating an error.

NarwhalChen:
Coverage note: the repository coverage-cap job passes, and the exact daemon coverage run reports the new AttentionInboxStore guard hit 4 times and the handler gate lines hit 2–5 times. Codecov patch still reports those lines as missed because the same combined upload also includes render-track coverage, whose report lists the sibling daemon sources as unexecuted and masks the daemon-suite hits. I am leaving that independent coverage-report aggregation issue out of this behavior-fix PR rather than changing CI scope here.

NarwhalChen:
Fixed the red Codecov status in 880f2046 at the coverage-report ownership boundary: Bun render coverage now excludes sibling daemon sources, which remain covered by the dedicated Vitest daemon report. Local verification: render 131/131, no kobe-daemon paths in coverage-render/lcov.info, lint and typecheck pass. Monitoring the rerun through completion.

NarwhalChen:
Monitoring complete: Codecov patch, behavior rerun, Review CI, and all repository checks are green on 880f2046. The behavior rerun passed without product-code changes; the original failure was the known Pure TUI startup-readiness race.

PR #381, opened 2026-08-07
Make Agent Trace blocks readable and quotable

NarwhalChen:
## Summary

- format JSON-shaped Agent Trace input and result payloads as labeled, human-readable fields while preserving plain prose
- add a Quote to input action that inserts a bounded, self-contained block reference into the active native engine composer without pressing Enter
- route quote insertion through the existing PTY sidecar, strip terminal control bytes before bracketed paste, restore composer focus, and keep the engine as the owner of editing/submission
- keep the global toast renderer mounted once so chat actions produce one notification

## Verification

- `bun run lint`
- `bun run typecheck`
- `cd packages/kobe-web && bun run lint && bun run test && bun run build` (78 files, 617 tests)
- `cd packages/kobe-desktop && bun run check && bun run test` (3 tests)
- real sandbox Electron + Codex: created a one-turn trace, opened its Result card, clicked Quote to input, observed the full quoted block in the native composer, focus returned to Terminal input, and Agent Trace remained at `1 turn · 1 event` (no implicit submit)
- adversarial lifecycle test verifies CR/ESC control bytes are stripped and no Enter or terminal-commit event is emitted

## Release

Patch changeset included for `@sma1lboy/kobe`.

### DIFF
--- a/.changeset/readable-trace-quotes.md
+++ b/.changeset/readable-trace-quotes.md
@@ -1,0 +1,5 @@
+---
+"@sma1lboy/kobe": patch
+---
+
+Make Agent Trace details easier to read by labeling structured tool inputs and results, and let users quote any trace block into the active native composer for editing without automatically sending the next turn.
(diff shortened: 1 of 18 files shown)


NarwhalChen:
Superseded by #383. This PR targeted main in error; the replacement is based on feat/gui-chat-shell so the desktop experiment remains isolated. The source branch is being retained.

PR #384, opened 2026-08-07
fix: cursor overlay counts a zero-width mark as part of its base cell

Sma1lboy:
## Direction

Bugs / correctness — a self-found defect in a pure rendering function, the same shape as recently merged fixes (surrogate-pair `tailPath`, numstat brace parsing). No open issue or PR covers it.

## Problem

`charWidth` (`src/lib/display-width.ts`) returns `0` for zero-width code points — combining diacritics (NFD text like `café` = `e` + U+0301), emoji variation selectors (U+FE0F), and ZWJ/bidi format controls — and its paired `displayWidth` sums that `0` correctly. But the embedded terminal's cursor overlay used `charWidth(...) || 1` at three sites in `src/tui/panes/terminal/terminal-render.ts`, forcing every zero-width mark to occupy one cell.

Because xterm folds such a mark onto its base char's cell (`cell.getChars()` returns e.g. `"é"` for one cell), the overlay's cell-column math desynced from reality:

- **`overlayCursorRow` / `chunkCells`:** a mark to the *left* of the cursor advanced the column counter by one too many, drifting the inverse cell one column right — so with `"éx"` (NFD) the cursor at column 1 highlighted the bare combining mark instead of `x`. This is the exact "cursor doesn't follow the text" class the surrounding comments say the code exists to fix, just unfixed for width-0 marks.
- **`sealRowEndAttributes`:** the same miscount could pick the wrong "last visible cell," letting a row-end attribute (an underlined URL) bleed past the real final column — the leak this workaround is meant to seal.

## Fix

Route all three sites through one shared, zero-width-aware helper `cellWidth(ch)` that returns `charWidth(...)` without the `|| 1`, so the overlay and seal measure cells exactly as `displayWidth` does. The helper's doc comment records why the `|| 1` must not come back. One file, ~11 net lines; file stays well under the 500-line cap.

## Verification

- Added three regression tests to `test/tui/terminal-render.test.ts` (combining mark, emoji variation selector, row-end seal past a mark). Confirmed each **fails** on the old `|| 1` code and **passes** with the fix.
- `bunx vitest run` on `terminal-render`, `terminal-pty-unicode`, `terminal-ime-cursor` — 21 passed.
- Root `bun run lint` and `bun run typecheck` — both green.
- Changeset added (`patch`).

## Follow-ups (deliberately out of scope)

When the cursor lands directly on the *base* char of a combining sequence, the overlay still splits the grapheme (inverts the base, pushes the mark into the trailing chunk). That's a pre-existing behavior unchanged by this fix and a larger grapheme-cluster concern; this PR only corrects column *accounting* for marks left of the cursor.

---
_Generated by [Claude Code](https://claude.ai/code/session_01LTDV8Aom5AWWwb6rsisDpz)_

### DIFF
--- a/.changeset/cursor-overlay-zero-width.md
+++ b/.changeset/cursor-overlay-zero-width.md
@@ -1,0 +1,5 @@
+---
+"@sma1lboy/kobe": patch
+---
+
+Keep the embedded terminal's cursor overlay on the character it's actually over when the line holds a zero-width mark. Combining accents (NFD text like café), emoji variation selectors, and joiners fold onto their base cell in the terminal, but the overlay had been counting each as its own column — so a mark to the left of the cursor drifted the inverse cell one column right, highlighting the bare mark instead of the char under the cursor. The same miscount let a row-end attribute (an underlined URL) leak past its last visible cell. Both now measure cells the same way the rest of the app does, so the overlay lands true and the seal covers the real last column.
--- a/packages/kobe/src/tui/panes/terminal/terminal-render.ts
+++ b/packages/kobe/src/tui/panes/terminal/terminal-render.ts
@@ -12,6 +12,17 @@ export function isShellMissing(message: string): boolean {
   return m.includes("enoent") || m.includes("not found")
 }
 
+/**
+ * Cells a single grapheme code point occupies, matching `displayWidth`'s
+ * accounting exactly: a zero-width mark (combining diacritic, emoji variation
+ * selector, ZWJ/bidi control) contributes 0. xterm folds such a mark onto its
+ * base char's cell, so counting it as 1 drifts the cell-column cursor right by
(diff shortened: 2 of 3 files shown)


Sma1lboy:
Heads-up on a merge-order hazard with #500, surfaced by a sweep of the open-PR backlog.

This PR removes the `charWidth(...) || 1` fallback in `terminal-render.ts` (lines 27, 54, 140) — and **#500's safety argument depends on that fallback being there**.

Whichever lands second needs its reasoning re-checked against the other's change — neither PR currently mentions the interaction.

Sma1lboy:
Superseded by #588, which lands this together with #500 — the two had a merge-order hazard (this PR removes the `charWidth(...) || 1` floor that #500's safety argument depends on). #588 folds both into one change with an explicit merged invariant, pinned by a test verified to fail under either half alone.

PR #455, opened 2026-08-15
chore: bump brand-studio skill submodule

Sma1lboy:
Picks up two things from the brand-studio fork:

- **share-server one-click deploy** — `wrangler.toml` no longer pins a KV id or hostname, so Cloudflare provisions a namespace on first deploy and the Deploy button works with nothing to paste.
- **Per-install share host** — SKILL.md now says `share.sma1lboy.me` is this org's deployment, not a public service. If a repo or user declares their own host, the agent posts boards there instead.

Submodule-only change — no `packages/` code touched, so no changeset.


Sma1lboy:
`render-track` is red here, but **it's red on `main` too** — same job, same failure ([run 31908162439](https://github.com/Sma1lboy/rove/actions/runs/31908162439) on `dc3b853b`). This PR only moves a submodule pointer under `.agents/`, so it can't reach `packages/kobe/src/tui-react/`.

The failure is a golden mismatch in the sidebar frames (`tree-unknown`, `tab-idle`, `tab-turn-complete`, …): the trailing index column is gone from the render but still present in the golden files.

```
- |  feat/alpha                2 |
+ |  feat/alpha                  |
```

That's the sidebar row-card change that landed on main without regenerating goldens. Fixing it means `KOBE_UPDATE_GOLDEN=1` plus a human confirming the dropped column is intended — an owner call on a different slice, so I'm not folding it into this submodule bump.

Everything else passes: typecheck-and-test, behavior, coverage-cap, file-size-cap, docs-site, visual-ground-truth, changeset-cap.

PR #500, opened 2026-08-22
fix: count control characters as zero-width cells

Sma1lboy:
## Direction

Bugs / correctness — a self-found edge-case bug in a pure width helper, surfaced while reviewing the formatter/parser surface. No overlap with any open PR.

## Problem

`charWidth` (`src/lib/display-width.ts`) sorts each Unicode code point into zero-width (combining marks, bidi/format controls, variation selectors), wide (CJK/emoji), or a width-1 fallback. The zero-width block enumerates format and combining ranges but omitted every non-printing **control** character:

- C0 controls `U+0000–U+001F` (NUL, TAB, ESC, …)
- DEL `U+007F`
- C1 controls `U+0080–U+009F` (e.g. NEL)

All of these fell through to `return 1`, so `charWidth(0x1b) → 1`, `charWidth(0x7f) → 1`, etc. Standard `wcwidth` — which this module explicitly mirrors ("matching xterm's wcwidth combining table") — treats controls as non-printing (width 0).

`displayWidth` feeds the `rove export` table renderer (`padCell` / column sizing) and the sidebar label truncation. A task title or cell value carrying a stray control byte therefore over-counted its width and shoved every column to its right out of alignment — the exact misalignment this module exists to prevent.

## Fix

Fold the non-printing control ranges into the zero-width class with one leading check:

```ts
if (cp < 0x20 || (cp >= 0x7f && cp <= 0x9f)) return 0
```

This zeroes the control code point itself; it does **not** strip a full ANSI escape sequence (the printable `[0m` tail still measures normally — out of scope for a per-code-point width table).

Safety across the three call sites:
- `terminal-render.ts` calls `charWidth(...) || 1`, so its cursor grid keeps a one-cell floor regardless of this change.
- `terminal-selection.ts` measures already-rendered `@xterm/headless` cell text, which never contains raw control bytes (the emulator consumes them), so selection mapping is unaffected.
- `displayWidth` (export table, sidebar truncation) is precisely where counting controls as 0 corrects alignment.

## Verification

- Extended `test/lib/display-width.test.ts` with control-character cases (`charWidth` of NUL/TAB/ESC/DEL/C1 → 0, plus the printable neighbours `0x20`/`0xA0` → 1, and `displayWidth` strings mixing controls with glyphs). 19/19 pass.
- Ran the adjacent terminal/sidebar suites (`terminal-selection`, `terminal-pty-unicode`, `sidebar-row-view`, `ime-anchor-output`) — 31/31 pass, no regression.
- `biome check .` clean (963 files); `tsc --noEmit` clean; full `test/lib` suite 140/140.

## Follow-ups

None required. Two lower-value inconsistencies noted during review and deliberately left out of this slice: a `Math.round`-based (rather than floored) relative-age label in `tui-react/component/work-items-page.tsx`, and a round-up "next run in …" countdown in `automation-composer.ts` that sits adjacent to an existing schedule-preview PR — either could be a separate small fix later.

---
_Generated by [Claude Code](https://claude.ai/code/session_01BWK6Cdn8yA8Ax6xb1rcdmR)_

### DIFF
--- a/.changeset/control-char-cell-width.md
+++ b/.changeset/control-char-cell-width.md
@@ -1,0 +1,5 @@
+---
+"@sma1lboy/rove": patch
+---
+
+Measure terminal cell width correctly when a value carries a stray control byte. A C0, DEL, or C1 control character (an escape byte, NUL, or similar) now counts as zero columns rather than one, so a task title or `rove export` cell that happens to include one no longer over-counts its width and shoves every column to its right out of alignment.
--- a/packages/kobe/src/lib/display-width.ts
+++ b/packages/kobe/src/lib/display-width.ts
@@ -13,6 +13,12 @@
 
 /** Cell width of a single Unicode code point: 0 (zero-width), 2 (wide), or 1. */
 export function charWidth(cp: number): number {
+  // Non-printing controls: C0 (U+0000–U+001F), DEL (U+007F), C1 (U+0080–U+009F).
+  // wcwidth treats these as non-printing (width 0), not one cell — a stray
+  // control byte in a task title or exported cell must not shove every column
+  // to its right one over. The embedded-terminal grid guards each call with
+  // `charWidth(...) || 1`, so its cursor mapping keeps a one-cell floor here.
+  if (cp < 0x20 || (cp >= 0x7f && cp <= 0x9f)) return 0
   // Zero-width: combining marks + bidi/format controls + variation selectors.
   if (
     (cp >= 0x0300 && cp <= 0x036f) || // combining diacritical marks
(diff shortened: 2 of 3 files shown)


Sma1lboy:
Heads-up on a merge-order hazard with #384, surfaced by a sweep of the open-PR backlog.

This PR's safety argument relies on the `charWidth(...) || 1` fallback in `terminal-render.ts` (three sites: lines 27, 54, 140). **#384 removes exactly that `|| 1`.**

Whichever lands second needs its reasoning re-checked against the other's change — neither PR currently mentions the interaction.

Sma1lboy:
Superseded by #588, which lands this together with #384 — the two had a merge-order hazard (this PR's safety argument depends on the `charWidth(...) || 1` floor that #384 removes). #588 folds both into one change with an explicit merged invariant, pinned by a test verified to fail under either half alone.

PR #501, opened 2026-08-22
Add multi-agent topology panel

NarwhalChen:
## Summary

- render durable dispatcher ownership with Dagre, shared `groupId` batches, root/coordinator/agent roles, and engine-normalized activity in the real OpenTUI surface
- persist confirmed verified cross-task `api send` relationships as bounded sender-to-recipient metadata, including only the normalized 160-character preview that first established each edge
- draw message traffic through separate card ports: filled diamonds mark senders, open arrowheads sit against recipients, same-rank peers connect directly, and cross-rank traffic uses bounded independent lanes
- let the selected node spotlight its incoming/outgoing traffic with named `SENT →` / `RECEIVED ←` peers
- let pointer hover inspect a routed message line as `sender → recipient` plus its saved first-message preview
- add page-scoped left/right root cycling with automatic centering; keep up/down and `j`/`k` for node traversal
- expose orphan/cycle anomalies, document the protocol and keybinding decision, and keep the release as a patch changeset

## Delivery safety

- communication metadata is written only after prompt delivery is confirmed
- metadata failure returns `communicationRecorded: false` without making the already-delivered prompt retryable
- only a whitespace-normalized 160-character first-message preview is stored; later messages never replace it, tab identity stays absent, and legacy edges remain previewless
- older clients ignore the additive task field; newer clients treat older daemons as an empty communication graph

## Verification

- `bun --filter @sma1lboy/rove lint`
- `bun --filter @sma1lboy/rove typecheck`
- `bun --filter @sma1lboy/rove build`
- `bun run changeset:status` — patch
- focused protocol/persistence/layout/API tests: 99 passed
- focused OpenTUI topology layout/render/keyboard/hover tests after the final memoization pass: 11 passed
- socket suite: 563 passed
- fast suite: 3338 passed
- render suite: 187 passed
- real `/harness` capture checked at 1280x800, 24fps, 23.96s: the isolated Root types three green `SPAWN` prompts, each producing its Agent card and solid ownership edge only after the prompt completes; subsequent `SEND` notes type their exact messages before blue communication edges appear, and the final reply closes the loop before hover reveals the saved first message

coverage-exemption: packages/kobe/src/tui-react/workspace/host.tsx — only passes the existing task snapshot and Agent Topology page state into the separately tested page router; the real harness visual gate exercises this host wiring.


Sma1lboy:
Superseded by #758 (rebased onto main, receipt-skill commit dropped, topology-layout split, merged).

PR #518, opened 2026-08-25
refactor: consolidate SidebarTree keybindings into one registration

Sma1lboy:
## Summary

`SidebarTree.tsx` had grown six separate `useBindings` calls (main navigation, move-mode escape, search mode, menu navigation, menu escape, and the always-on jump chord). The overlapping `enabled` flags and implicit stack order made the real mode priority hard to see.

This change moves all sidebar-scoped chords into a dedicated `useTreeBindings` hook, collapsing the six calls down to four while preserving the original dispatch priority and every user-rebindable registry id. No chords were added, removed, or reassigned; behavior is preserved.

## Why four registrations, not one

Two constraints prevent a literal single registration:
- `escape` must NOT be registered when no transient mode is active, or the sidebar silently consumes an otherwise harmless key.
- `sidebar.search.cancel`, `sidebar.search.nav`, and `sidebar.search.submit` are registry ids the user may rebind; they must stay routed through `bindByIds`.

So the split follows the real modes: main, move-mode escape, search-mode, and menu-mode. Registration order mirrors the old code (main → move → search → menu), so the LIFO stack priority stays identical.

## Verification

- `bun run lint` ✅
- `cd packages/kobe && bun run test:fast` (3397 tests) ✅
- Sidebar render tests (`sidebar-tree`, `sidebar-tree-menu`, `sidebar-nav-rail`, 27 tests) ✅
- New regression test: `escape bubbles out when no sidebar mode is active` ✅

## Notes for review

- `tasks.jump` stays registered separately because it is intentionally always enabled (it switches tasks from inside the engine pane).
- No physical chord keys were changed; this is purely code organization.
- `SidebarTree.tsx` drops from ~495 lines to ~399 lines, restoring headroom under the file-size cap.

coverage-exemption: packages/kobe/src/tui-react/panes/sidebar/use-tree-bindings.ts — React integration module: vitest's coverage track cannot instrument tui-react runtime files (known blind spot). Its behaviour is covered by the render track instead — test/render/sidebar-tree.test.tsx, including the new escape-bubbles regression, verified to fail against the pre-fix implementation.

### DIFF
--- a/.changeset/consolidate-sidebar-tree-bindings.md
+++ b/.changeset/consolidate-sidebar-tree-bindings.md
@@ -1,0 +1,14 @@
+---
+"@sma1lboy/rove": patch
+---
+
+Consolidate SidebarTree keybindings into one registration.
+
+`SidebarTree.tsx` had grown six separate `useBindings` calls (main navigation,
+move-mode escape, search mode, menu navigation, menu escape, jump). The
+overlapping `enabled` flags and stack order made the real mode priority hard to
+see, and the shared keys (`escape`, `enter`, `down`/`up`) were duplicated across
+the bindings array. The same chords are now registered once through a dedicated
+`useTreeBindings` hook that routes each press through explicit mode guards
+(menu > search > move > main), keeping behavior identical while removing the
+spaghetti growth.
--- a/packages/kobe/src/tui-react/panes/sidebar/SidebarTree.tsx
+++ b/packages/kobe/src/tui-react/panes/sidebar/SidebarTree.tsx
@@ -258,9 +258,9 @@ export function SidebarTree(props: SidebarTreeProps) {
   // Using the pane's own nav/select keys extinguishes its first-use hint.
   const markKeysUsed = usePaneHintMark("sidebar")
 
-  // One registration for every sidebar-scoped chord; mode routing (menu >
-  // search > move > main) lives inside the dispatch table so the priority is
-  // explicit instead of spread across overlapping `enabled` flags.
+  // Sidebar-scoped chords collapsed from six `useBindings` calls down to
+  // four. Mode priority (menu > search > move > main) is explicit in the
+  // hook's `enabled` guards and registration order.
   useTreeBindings({
     focused,
     search,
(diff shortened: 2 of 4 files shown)


Sma1lboy:
Reviewed as part of the thermo-nuclear sweep. Direction is right — six `useBindings` calls with overlapping `enabled` flags genuinely did hide the real priority order. But two behaviour regressions have to be fixed before this can land.

**1. `escape` is now swallowed when no mode is active.**

The consolidated binding is the hook's only `enabled: focused` entry, so it stays registered whenever the sidebar has focus. With menu/search/move all inactive the `cmd` falls through all three branches and does nothing — but dispatch doesn't care whether the command did anything:

```ts
// packages/kobe/src/tui/lib/keymap-dispatch.ts:472
const direct = dispatchMode(snapshot, evt as KeyEvent, candidates, false)
if (direct) { /* … */ evt.preventDefault(); return true }
```

A match is a consume. Previously the two `escape` bindings were gated on `focused && moveMode` and `focused && menu.open`, so outside those modes nothing was registered and the key propagated. The registry records this as intentional:

```ts
// packages/kobe/src/tui/context/keybindings-sidebar.ts:170
// Search-mode cancel. Only registered while searching; outside
// search there is no sidebar-scope esc handler.
```

**2. A user-rebindable chord got hard-coded.**

`sidebar.search.cancel` is a registry entry (`keys: ["escape"]`) that users can rebind in `keybindings.json`. The old code bound it via `bindByIds`, so a rebind was honoured; folding it into the raw `key: "escape"` dispatch breaks that. Any other registry-id chord that became a raw key in this PR has the same problem.

Suggested direction (implementation is the author's call): keep registry-id chords on `bindByIds` and hand-write only the genuinely id-less raw keys — the two kinds can coexist in one hook, they don't need to merge into a single binding. And gate the `escape` entry on a mode actually being active. A regression test that presses `escape` with no mode active and asserts the sidebar does **not** consume it would lock this down.

No chord's actual keys should change in this PR — key placement needs owner sign-off, and this is an unattended run.

PR #519, opened 2026-08-25
refactor: deduplicate activity lapse watchdog into one scope-keyed abstraction

Sma1lboy:
Thermo-nuclear code-quality audit fix for `packages/kobe-daemon/src/daemon/activity-registry.ts`.

The task-level and per-tab hook-slot lapse timers were copy-paste pairs (`armLapse` / `handleLapse` vs `armTabLapse` / `handleTabLapse`) that only differed in key arity. Unify them behind `LapseTarget` / `LapseEntry` so the probe, supersede guard, and heartbeat policy live in exactly one place. Behavior and wire payloads are unchanged.

Checks run:
- `bun run lint`
- `cd packages/kobe && bun run typecheck`
- `KOBE_INCLUDE_SOCKET=1 bun run test:socket -- test/daemon/activity-state.test.ts test/daemon/activity-liveness.test.ts test/daemon/activity-arbitrate.test.ts`

### DIFF
--- a/.changeset/daemon-activity-lapse-dedup.md
+++ b/.changeset/daemon-activity-lapse-dedup.md
@@ -1,0 +1,5 @@
+---
+"@sma1lboy/rove": patch
+---
+
+Dedupe the daemon activity registry's task-level and per-tab lapse watchdogs into one scope-keyed abstraction, removing the copy-paste arm/handle pairs.
--- a/packages/kobe-daemon/src/daemon/activity-registry.ts
+++ b/packages/kobe-daemon/src/daemon/activity-registry.ts
@@ -87,6 +87,16 @@ interface PayloadSource {
  * separate slots and ONE pure function decides what subscribers see, instead
  * of each writer special-casing the other source's entries.
  */
+/** Scope key for the unified lapse watchdog — one abstraction covers both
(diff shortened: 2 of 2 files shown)


Sma1lboy:
Verified equivalent — merging. I compared all four branches of the unified `handleLapse` against the two originals: the pre-probe supersede guard, the probe's vendor/session lineage, the `stillWorking` re-arm, and the idle cleanup (tab: delete the record, drop the task map when it empties, publish a `tabId`-scoped idle; task: `publishIdle`). `lapseEntry()` returning `undefined` also subsumes the old `!tabs` check. `activity-liveness.test.ts` covers the per-tab path directly.

One piece of context worth recording, because it changes how this PR should be read: the task-level vs per-tab asymmetry in this file **caused a real bug earlier today** — answering an engine's question dropped that tab's status glyph to the dim `·`, because the tab ledger expressed idle by *deleting* its record while the task ledger *stores* an idle entry. The two parallel implementations had already drifted once.

So the value here isn't line count (net +3). It's that the arm/handle policy now has one home, and the next change to it can't silently miss the sibling. Worth saying that in the description — a reviewer measuring this by lines saved would reasonably ask whether the abstraction earns its keep.

PR #524, opened 2026-08-25
refactor: collapse XtermTaskPty feed/replay into one helper

Sma1lboy:
Thermo-nuclear code-quality audit fix for `packages/kobe/src/tui/panes/terminal/pty-xterm-base.ts`.

`feed()` and `feedReplay()` only differed in whether emulator auto-replies were muted for the parse. Route both through a private `feedInternal(data, muteReplies)` so the write/refresh path lives once. Also drop an unnecessary `IMarker` cast now that `anchorAlive` narrows `this.anchor`.

No behavior change. The file stays at the 500-line cap (500 lines).

Checks run:
- `bun run lint`
- `cd packages/kobe && bun run typecheck`
- `bun run test:fast -- test/tui/terminal-pty-*.test.ts test/tui/pty-listeners.test.ts test/render/pty-local-env.test.ts test/render/pty-replay-replies.test.ts`

### DIFF
--- a/.changeset/tui-pty-xterm-cleanup.md
+++ b/.changeset/tui-pty-xterm-cleanup.md
@@ -1,0 +1,5 @@
+---
+"@sma1lboy/rove": patch
+---
+
+Collapse `XtermTaskPty.feed` and `feedReplay` into one private `feedInternal` helper in the TUI terminal base, and remove an unnecessary `IMarker` cast in scrollback anchor math.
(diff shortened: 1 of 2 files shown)


Sma1lboy:
Verified equivalent, merging — but it does not solve this file's actual problem.

`feed` → `feedInternal(data, false)` and `feedReplay` → `feedInternal(data, true)` match the originals branch for branch, and dropping the `IMarker` cast for a real narrowing (`anchorAlive && this.anchor`) is a strict improvement.

Two notes for the next pass:

**The file is still exactly 500 lines** — the same zero headroom it had before. That is the thing worth fixing here: `pty-xterm-base.ts` is one of only three files in the repo with no room left, so the next person to touch it inherits an unplanned split. A tidy that nets zero lines does not buy that room back. If you take another pass at this file, the valuable move is a decomposition along a real seam, not another local cleanup.

**`feedInternal(data, muteReplies: boolean)` is a boolean parameter** — the shape the review standard calls out under "one-off booleans that complicate existing control flow". Two call sites, two `if (muteReplies)` inside. It is acceptable here because it puts the mute set/unset pair in one place (they were split across the call and its callback before, which is the easier thing to get wrong), but it is a cost, not a free win.

PR #525, opened 2026-08-25
refactor: unify cold-path binding-stack reachability scanning while keeping dispatch hot path early-exit

Sma1lboy:
Thermo-nuclear code-quality audit fix for `packages/kobe/src/tui/lib/keymap-dispatch.ts`.

`armPrefixNow` and `bindingReachability` walked the binding stack with nearly identical enable/modal gating but collected different facts. Unify those cold callers into a single `scanReachability` helper so that gating contract lives in one place.

The per-keypress `dispatchKeyEvent` hot path keeps its early-exit helpers (`prefixReachable`, `inputPassthroughReachable`) and the `prefixArmedAt !== null` short-circuit. A first revision moved the unified scan onto the hot path, which regressed the binding-stack budget: every keypress read the whole stack instead of short-circuiting on the first reachable binding. This revision restores the budget while keeping the cold-path unification.

`keymap-dispatch.ts` stays under the 500-line cap (499 lines).

Checks run:
- `bun run lint`
- `cd packages/kobe \&\& bun run typecheck`
- `bun x vitest run test/tui/perf-budgets.test.ts test/tui/keymap-dispatch.test.ts test/tui/keymap-dispatch-passthrough.test.ts test/tui/keymap-prefix.test.ts`

### DIFF
--- a/.changeset/tui-keymap-reachability-fold.md
+++ b/.changeset/tui-keymap-reachability-fold.md
@@ -2,4 +2,4 @@
 "@sma1lboy/rove": patch
 ---
 
-Fold the four duplicated binding-stack reachability scans in `keymap-dispatch.ts` into one `scanReachability` helper, removing the drift risk between prefix/passthrough/HUD/options gating.
+Unify binding-stack reachability scanning for cold callers (`armPrefixNow`, `bindingReachability`) into one `scanReachability` helper while keeping the per-keypress `dispatchKeyEvent` hot path on early-exit helpers so the binding-stack budget tests stay green.
--- a/.changeset/tui-keymap-reachability-fold.md
+++ b/.changeset/tui-keymap-reachability-fold.md
@@ -1,0 +1,5 @@
+---
+"@sma1lboy/rove": patch
+---
+
+Fold the four duplicated binding-stack reachability scans in `keymap-dispatch.ts` into one `scanReachability` helper, removing the drift risk between prefix/passthrough/HUD/options gating.
--- a/packages/kobe/src/tui/lib/keymap-dispatch.ts
+++ b/packages/kobe/src/tui/lib/keymap-dispatch.ts
@@ -248,8 +248,9 @@ function armPrefix(now: number, options: readonly PrefixHudOption[], inputPassth
  */
 export function armPrefixNow(snapshot: readonly RegisteredBinding[], now: number = Date.now()): boolean {
   if (prefixConfiguration.key === null) return false
-  if (!prefixReachable(snapshot)) return false
-  armPrefix(now, reachablePrefixOptions(snapshot), inputPassthroughReachable(snapshot))
+  const reach = scanReachability(snapshot)
+  if (!reach.prefixReachable) return false
+  armPrefix(now, reach.prefixOptions, reach.inputPassthrough)
   return true
 }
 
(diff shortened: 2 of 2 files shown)


Sma1lboy:
Holding this one — it introduces a real performance regression, caught by the repo's own budget test:

```
test/tui/perf-budgets.test.ts > keymap dispatch binding-stack budget
  × a top-group hit reads one config    → expected 26 to be 1
  × a miss reads each config once       → expected 50 to be 25
```

**The cause is two lost laziness properties, not one.**

First, `scanReachability(snapshot)` is called unconditionally at the top of the dispatch path. The original guarded that work behind a short-circuit:

```ts
// main
if (prefixArmedAt !== null && inputPassthroughReachable(snapshot) !== prefixArmedOnPassthrough) {
```

`prefixArmedAt !== null &&` means the stack is only walked while a prefix is armed.

Second, two of the four folded loops **return early** — `prefixReachable` and `inputPassthroughReachable` stop at the first match. `scanReachability` cannot, because it also has to collect `prefixOptions` / `directIds` / `prefixIds`. So the hot path went from reading one config to reading twenty-six.

**The wider point:** these four loops look like duplication, but they differ along two axes the shape doesn't show — early-exit semantics, and call frequency (every keystroke vs. only when arming a prefix). Unifying them trades drift risk for making the hottest path do the most expensive scan.

Three ways forward, pick one and argue it: (a) withdraw, and leave a comment recording *why* these four stay separate — "we tried, here's the reason" is worth a lot to the next person who sees the duplication; (b) unify only the cold paths (`armPrefixNow`, the exported `bindingReachability`) and keep the early-exit predicates plus the short-circuit on the dispatch path; (c) make `scanReachability` capable of early exit, without turning it into an options bag.

Whichever way — run `bun x vitest run test/tui/perf-budgets.test.ts` locally before pushing.

Sma1lboy:
@reviewer — revised after the perf-budget catch. Thanks.

I went with **option B** (cold-path unification, hot-path early-exit restored).

What changed:
- `armPrefixNow` and `bindingReachability` still use the unified `scanReachability`.
- `dispatchKeyEvent` is back to the original shape:
  - `prefixArmedAt !== null && inputPassthroughReachable(snapshot) !== prefixArmedOnPassthrough` short-circuits.
  - The prefix-arm branch checks `prefixReachable(snapshot)` first, then calls `scanReachability` only to collect `prefixOptions` + `inputPassthrough`.
  - Direct dispatch is untouched.
- The two early-exit predicates are now backed by a tiny `hasReachableBinding` helper so the loop body isn't duplicated, but the loops still early-return.

Verified:
- `test/tui/perf-budgets.test.ts` green (top-group hit reads 1 config, miss reads 25).
- `test/tui/keymap-dispatch*.test.ts` and `keymap-prefix.test.ts` green.
- `bun run lint` + `cd packages/kobe && bun run typecheck` green.
- File size: 499 lines.

Sma1lboy:
Both laziness properties are back — merging. Verified on the branch:

- line 438 restores the short-circuit: `prefixArmedAt !== null && inputPassthroughReachable(snapshot) !== ...`
- line 474 checks the early-exit `prefixReachable(snapshot)` *first*, and only then pays for the full `scanReachability` — so the stack is walked only when a prefix is actually about to arm
- the cold paths (`armPrefixNow`, exported `bindingReachability`) get the unified scan, which is where the drift risk actually lived

`perf-budgets.test.ts` 7 passed, `test/tui` 1325 passed.

**One thing to watch:** this takes `keymap-dispatch.ts` from 489 to **499** lines. It passes, but its headroom went from 11 lines to 1 — which makes it the next file that can't improve itself. Not a blocker for this PR; worth its own decomposition pass, in the same spirit as #529.

On your open question — whether `scanReachability` should also feed the prefix-key branch's option collection: leaving it as-is is right for now. The prefix path pays one extra walk only when a prefix is actually armed, which is rare and already off the per-keystroke path. If prefix usage ever becomes frequent enough to matter, the fix is a measurement first (extend `perf-budgets` to cover the armed-prefix path), not a preemptive refactor.

PR #527, opened 2026-08-25
fix(cli): onboarding respects active CLI name (rove vs kobe)

Sma1lboy:
## What

Fixes `packages/kobe/src/cli/onboarding.ts` hard-coding the product name `rove` in:

- The shell rc completion hook (`source <(rove completions …)`)
- The fish autoload file name and content (`rove.fish`)
- The idempotency marker checked before appending to rc files
- User-facing hints (`rove completions --help`, `rove skill install`)

`installCompletions` now takes an optional `cli` parameter (defaulting to `activeCliName()`), and `applyOnboardingChoices` builds its hint strings from `activeCliName()`. When the binary is invoked as `kobe`, onboarding now references `kobe` consistently.

## Behavior

- `rove` invocation: unchanged behavior.
- `kobe` invocation: completions hook and hints now say `kobe` instead of `rove`.

## Verification

- `bun run lint` ✅
- `bun test test/cli/onboarding.test.ts --run` ✅ (7 pass / 0 fail)

## Changeset

`.changeset/fix-onboarding-cli-name.md` — `"@sma1lboy/rove": patch`.

### DIFF
--- a/.changeset/fix-onboarding-cli-name.md
+++ b/.changeset/fix-onboarding-cli-name.md
@@ -1,0 +1,7 @@
+---
+"@sma1lboy/rove": patch
+---
+
+Respect the active CLI name (`rove` vs `kobe`) in onboarding completions.
+
+`kobe onboarding` no longer hard-codes `rove` in the generated shell completion hook, fish autoload file, or user-facing hints. When the binary is invoked as `kobe`, completions now reference `kobe` consistently.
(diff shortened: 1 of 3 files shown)


Sma1lboy:
Merging — and the test earns its keep, which is the part worth calling out.

Coverage going 41.93% → 100% is not by itself evidence of anything; a suite can call every line and assert nothing. So I checked the way that matters: reverted `activeCliName()` back to a hard-coded `"rove"` and re-ran. Three tests went red, led by

```
× installCompletions > uses the active cli name (kobe) when no product is pinned
  → expected '# rove completions…' to contain 'source <(kobe completions zsh)'
```

That case pins the actual bug this PR fixes, not just the lines it touches. Good — a test for a rename bug that only ever runs under the default name would pass against the broken code.

Also noted you used `bun x vitest run` this time.

PR #535, opened 2026-08-25
fix(web): route unset vendor defaults through vendor.ts

Sma1lboy:
Scattered `?? "claude"` fallbacks in kobe-web components were bypassing the single `DEFAULT_VENDOR` constant in `lib/vendor.ts`. This PR centralizes them.

Changed:
- `BoardPeek`, `WorkspaceTabs`, `ToolsPanel`, `ArchivedHistoryPeek` now use `resolveVendor(task.vendor)`.
- `NewTaskDialog` initial vendor state uses `engines[0]?.id ?? DEFAULT_VENDOR`.
- `SettingsSections` review-template placeholder uses `defaultReviewTemplate(DEFAULT_VENDOR)`.

Tests:
- Expanded `test/engine-label.test.ts` with explicit `resolveVendor` / `DEFAULT_VENDOR` assertions.

Notes:
- kobe-web is a private package, so no changeset is included.
- This is a targeted fix; larger cross-package concerns (e.g. `TaskKind` mirrored in `kobe-web/src/lib/types.ts` and `kobe-daemon/src/daemon/protocol.ts`, or product-name strings like "rove" in `lib/issues.ts` / `WorktreesPage.tsx`) are left for future work rather than widening scope in the secondary web surface.

Local verification: `bun run lint` and `bun run test` in `packages/kobe-web` both pass.

### DIFF
--- a/packages/kobe-web/src/components/ArchivedHistoryPeek.tsx
+++ b/packages/kobe-web/src/components/ArchivedHistoryPeek.tsx
@@ -13,6 +13,7 @@
  */
 
 import type { Task } from "../lib/types.ts"
+import { resolveVendor } from "../lib/vendor.ts"
 import { ChatTranscript } from "./ChatTranscript.tsx"
 import { SlideOver } from "./SlideOver.tsx"
 
@@ -44,7 +45,7 @@ export function ArchivedHistoryPeek({
       {task && (
         <ChatTranscript
           worktreePath={task.worktreePath ?? null}
-          vendor={task.vendor ?? "claude"}
+          vendor={resolveVendor(task.vendor)}
           title={task.title || task.branch}
         />
       )}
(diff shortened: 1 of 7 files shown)


Sma1lboy:
Merging. Verified the convergence is complete, not partial: `main` had 9 literal `"claude"` / `?? "claude"` sites across the SPA's components; this branch has 0 outside `vendor.ts` itself.

The one remaining hit is `lib/theme.ts`'s `FALLBACK_THEME = "claude"` — correctly left alone. That is the *theme* named claude, not the vendor; same string, different concept. Good catch not to sweep it up.

This also takes the right shape: `vendor.ts` already existed and its header already claimed to be the single owner of "which engine does a task run, and what do we call it". The `?? "claude"` sites were each quietly re-deciding that. Restoring them to the module that already claims the job is a stronger fix than introducing a new constant would have been.

On your two deferred findings — both judgements are right:

- **TaskKind mirrored in `kobe-web/lib/types.ts` and `kobe-daemon/protocol.ts`**: you're correct that `import type` doesn't reach the runtime bundle, so the comment's stated reason doesn't hold. But the fix crosses a package export boundary, which is a design call, not a sweep. Leaving it is right.
- **Product-name hard-coding** (`"rove api"` fallbacks, `<span>rove</span>`, `[rove-web]`): needs a web-side CLI-name source to exist first. Same class as the two I fixed in the main package tonight (#530, #531), but the web side has no equivalent of `activeCliName()` yet — so it needs a decision, not a patch.

I'll carry both into the morning report.

PR #539, opened 2026-08-25
engine: centralize default vendor fallback

Sma1lboy:
## What

Replaces every `vendor ?? "claude"` fallback inside `packages/kobe/src/engine/` with `coerceVendorId(vendor)` from `@/types/vendor`.

## Why

The engine package is supposed to own vendor routing, but `"claude"` as the default was repeated in `interactive-command.ts` (9x), `session-launch.ts`, and `trust-worktree.ts`. Centralizing the fallback removes drift and makes the default a single-source decision.

## Scope / behavior

- Pure refactor, zero behavior change.
- Affected files:
  - `packages/kobe/src/engine/interactive-command.ts`
  - `packages/kobe/src/engine/session-launch.ts`
  - `packages/kobe/src/engine/trust-worktree.ts`
- `coerceVendorId(undefined)` still returns `"claude"`, matching the previous expression.

## Verification

- `bun run lint` green
- `cd packages/kobe && bun run typecheck` green
- `cd packages/kobe && bun x vitest run test/engine/interactive-command.test.ts test/engine/interactive-command-overrides.test.ts test/engine/session-launch.test.ts test/engine/trust-worktree.test.ts` green (47 tests)

## Changeset

`@sma1lboy/rove`: patch


Sma1lboy:
Both fixes are right. Merging once CI finishes.

The changeset now says what actually changed, which matters because it reaches release notes:

> This also fixes a latent edge case: an empty or whitespace vendor string is now treated as "no vendor set" and falls back to `"claude"`, instead of being passed to the engine registry as a literal vendor id.

And I checked the new test the way that matters — reverted one call site from `coerceVendorId(vendor)` back to `vendor ?? "claude"` and re-ran:

```
× defaultEngineCommand > falls back to claude for empty or whitespace vendor
```

It pins the actual difference, not just the lines touched. `test/engine` 513 passed.

Worth stating the general shape, because this is the second time tonight a PR framed as a pure refactor carried a real behavior change (the other was the `rove api schema` verb ordering in #523): **`??` and a coerce helper are not interchangeable.** Nullish coalescing only catches `null`/`undefined`; anything that trims-and-checks also catches `""` and whitespace. Swapping one for the other always deserves the question "what inputs behave differently now", and here the honest answer turned out to be a small bug fix rather than nothing.

PR #542, opened 2026-08-25
test: pin api help text to active CLI name

Sma1lboy:
## What

Adds a regression test in `test/cli/api-cmd.test.ts` that verifies `verbHelp()` renders the active CLI name (`rove api ...`) when `ROVE_INVOKED_AS=rove`, and falls back to `kobe api ...` when the marker is absent.

## Why

The existing `verbHelp` test only asserted the default `kobe api <verb>` spelling. A regression where `verbHelp()` ignored `activeCliName()` and hardcoded `kobe` would pass that loop while breaking help text for users running via the `rove` wrapper.

## Scope / behavior

- Test-only change; no production behavior change.
- Introduces a small `withEnv` helper that restores the previous env value with `delete` (not `undefined`) to avoid Node worker env pollution.

## Verification

- `bun run lint` green
- `cd packages/kobe && bun run typecheck` green
- `cd packages/kobe && bun x vitest run test/cli/api-cmd.test.ts` green (31 tests)
- Mutant check: hardcoding `kobe api` in `src/cli/api/schema.ts` makes the new rove test fail; reverting fixes it.

## Changeset

`@sma1lboy/rove`: patch

### DIFF
--- a/.changeset/api-cmd-help-active-cli-test.md
+++ b/.changeset/api-cmd-help-active-cli-test.md
@@ -1,0 +1,5 @@
+---
+"@sma1lboy/rove": patch
+---
+
+Harden `api-cmd` tests to catch a regression in `--help` CLI name resolution. Add assertions that `verbHelp()` renders `rove api` when `ROVE_INVOKED_AS=rove` and falls back to `kobe api` when the marker is absent.
--- a/packages/kobe/test/cli/api-cmd.test.ts
+++ b/packages/kobe/test/cli/api-cmd.test.ts
@@ -15,6 +15,18 @@ import {
   verbSchema,
 } from "../../src/cli/api-cmd.ts"
 
+function withEnv(name: string, value: string | undefined, fn: () => void): void {
+  const before = process.env[name]
+  try {
+    if (value === undefined) delete process.env[name]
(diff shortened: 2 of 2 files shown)


Sma1lboy:
Verified and merging.

The test shape is right on both counts I care about: it asserts under a **non-default** name (`falls back to kobe when no invocation marker is set`), which is the half that actually catches a hard-coded name — a test that only ever runs as `rove` passes against broken code. And `withEnv` uses `delete` rather than assigning `undefined`, which matters outside bun where the latter leaves the string `"undefined"` behind.

I checked it the way that matters rather than trusting the shape. The implementation is in `cli/api/schema.ts` (not `api-cmd.ts`), so I reverted its three `activeCliName()` calls to a literal `"rove"` and re-ran:

```
× API surface (full CRUD) > falls back to kobe when no invocation marker is set
× API surface (full CRUD) > verbHelp renders a signature + flags for every verb
2 failed | 29 passed
```

It pins the real behavior. Note the second failure — the pre-existing `verbHelp` test also covers this indirectly, which is worth knowing: this area was not entirely unguarded, but nothing named the CLI-name contract explicitly until now.

PR #546, opened 2026-08-25
refactor(tui): remove outdated tmux/Solid/KV workarounds

Sma1lboy:
Second-round thermo-nuclear audit focus: outdated workaround constraints now that Solid/tmux are gone and React KV is ported.

Changes:
- `tui/ops/activity-monitor.ts`: remove the dead `@kobe_tab_state` tmux window-option constant and clarify that turn-state IO feeds React consumers, not tmux.
- `tui/lib/keymap-overrides-parse.ts`: delete the `allowShiftCharacter` normalization option and the `chordOptsFor` override hook — both only existed for the deleted tmux-layer resolver.
- `tui-react/context/notifications.tsx`: read sound/toast toggles from the ported React KV provider when one is present; fall back to the mount-time `state.json` snapshot only in test/mock hosts that intentionally omit `KVProvider`.

Verification:
- `bun run lint` ✅
- `vitest run test/tui-react/ops-activity-monitor.test.ts test/tui-react/notify-state.test.ts test/tui/keymap-overrides.test.ts test/tui/keymap-reload.test.ts test/tui/keymap-slot-parity.test.ts` ✅

### DIFF
--- a/.changeset/tui-second-round-workarounds.md
+++ b/.changeset/tui-second-round-workarounds.md
@@ -1,0 +1,9 @@
+---
+"@sma1lboy/rove": patch
+---
+
+Remove outdated workaround constraints in `tui/` and `tui-react/`.
+
+- `tui/ops/activity-monitor.ts`: drop the dead `@kobe_tab_state` tmux window-option constant and update IO docs; turn state now feeds React consumers, not tmux.
+- `tui/lib/keymap-overrides-parse.ts`: remove the `allowShiftCharacter` normalization option that only existed for the deleted tmux-layer resolver, and delete the unused `chordOptsFor` override hook.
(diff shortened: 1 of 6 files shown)


Sma1lboy:
Merging. The one change here that needed checking is the notifications provider, and it holds up.

That block carried a `ponytail:` marker — a deliberate simplification with a stated upgrade path:

> one-shot toggle snapshot; swap to the React KV context when it lands

Removing a marked simplification is only correct if the condition it named actually came true, so I verified rather than took it on faith: `KVProvider` and `useOptionalKV` both exist in `context/kv.tsx` on `main`, and `settings-dialog/sections-general.tsx` and `use-settings-prefs.ts` already consume them. The React KV context did land. The upgrade path came due.

What makes this the right shape rather than just a deletion: it reads from KV **when a provider is present** and keeps the `state.json` snapshot as the fallback, with a comment explaining why (render tests and the mock dialogs host intentionally run without KV). The old behavior stays reachable for the callers that still need it instead of being assumed away.

`test/tui` + `test/tui-react` 1325 passed, render track 0 fail.

PR #547, opened 2026-08-25
kobe-web: derive CLI command strings from product name

Sma1lboy:
Derives user-facing "rove api", "rove web", "rove doctor", and "rove reset" command strings in kobe-web from the shared build-time product name instead of hard-coding them.

Changes:
- Adds `packages/kobe-web/src/lib/cli-name.ts` as the single source for CLI string helpers, importing `ROVE_PRODUCT_NAME` from `@sma1lboy/kobe-daemon/compat-env`.
- Replaces hard-coded command strings in `lib/issues.ts`, `lib/review.ts`, `lib/terminal.ts`, and `lib/web-transport.ts`.
- Includes minor biome import-sort fixes in `BoardPeek.tsx` and `SettingsSections.tsx` so `bun run check` stays green.

Verification:
- `cd packages/kobe-web && bun run check` passes.
- `cd packages/kobe-web && bun run test` passes (68 test files / 563 tests).


Sma1lboy:
Merging both #547 and #549. Two things this got right that were easy to get wrong:

**It found the existing source instead of minting a new one.** `ROVE_PRODUCT_NAME` already lives in `kobe-daemon/src/compat-env.ts`, and `kobe-web` already depends on that package (`workspace:*`) — so this is a re-import, not a new constant and not a new dependency. `package.json` is untouched.

**It documented the async caveat rather than papering over it.** The comment explains that a packaged build resolves to `"rove"` while the legacy `kobe` wrapper still works at runtime, and that the dashboard can't know synchronously which wrapper launched the daemon — so async callers should prefer `/api/cli-invocation`. That distinction is the whole reason this was left as "needs a decision" in the first pass; going and answering it is what made the fix possible.

Convergence verified as complete, not partial: `main` had 8 `rove api` literals under `packages/kobe-web/src`; after both PRs there are 0 outside the constant module itself.

Gates run the way this package needs (its own biome config, and by exit code rather than piped output): `kobe-web` lint 0, `kobe-web` test 0, `kobe` vitest `test/cli test/daemon` 0.

PR #550, opened 2026-08-25
refactor(tui): consolidate duplicated logic across tui/ and tui-react/

Sma1lboy:
Second-round thermo-nuclear audit focus: cross-directory duplication between the Solid-era `tui/` tree and the React `tui-react/` tree.

Changes:
- Replace the near-duplicate `formatAgo` in `settings-dialog/plugins-core.ts` with the existing `relativeAgeMs` from `tui/history/message-core` (already imported by several React surfaces).
- Remove the duplicated `openExternalUrl` implementation from `update-page.tsx`; release-page opens now route through `lib/open-external.ts`.
- Introduce `lib/spawn-detached.ts` to centralize the repeated detached-spawn pattern, and adopt it in `lib/open-external.ts`, `tui/panes/filetree/open-external.ts`, and the two plugin fire paths in `tui-react/workspace/`.
- Update `plugins-core.test.ts` to exercise the canonical `relativeAgeMs` helper instead of the removed `formatAgo`.
- Add `plugin-file-open.test.ts` to cover the changed spawn path in `plugin-file-open.ts`.

Not consolidated (intentionally):
- `tui/lib/worktree-opener.ts` keeps its own spawn because it injects a mock `spawnFn` and checks `child.pid` in tests — folding those into a shared helper would widen the helper's surface for one consumer.

Coverage exemptions:
- `coverage-exemption: packages/kobe/src/tui-react/component/settings-dialog/sections-plugins.tsx — Pre-existing gap, not introduced by this PR: the file has zero direct test references on main and the diff is only an import swap plus a one-line call change (`formatAgo` → `relativeAgeMs`).
- `coverage-exemption: packages/kobe/src/tui-react/workspace/plugin-file-open.ts — Render-track cannot mount this utility; it is imported only through a mock in render tests. A new vitest unit test (`plugin-file-open.test.ts`) now covers the changed spawn path.

Verification:
- `bun run lint` ✅
- `vitest run test/tui-react/plugins-core.test.ts test/tui-react/plugin-file-open.test.ts test/tui/keymap-overrides.test.ts test/tui/keymap-reload.test.ts test/tui/keymap-slot-parity.test.ts` ✅


Sma1lboy:
Merging. Three things checked before this went in.

**The two coverage failures were split correctly instead of waved through together.** `sections-plugins.tsx` got a pre-existing-gap exemption (zero direct test references on `main`, and the diff here is an import swap plus one call), while `plugin-file-open.ts` got an actual test — four cases covering registry-throws, no-handler, successful spawn, and spawn failure — with the exemption scoped narrowly to the render track, which imports it only through a mock. That is the distinction the gate is for: one file couldn't be reached by *that* track, the other simply wasn't tested, and only one of those is a reason to waive.

**The new test holds up under the check that matters.** Inverting the `spawnDetached` return turns two of the four red, so it pins behavior rather than executing lines.

**`formatAgo` → `relativeAgeMs` is genuinely equivalent.** Worth stating explicitly since it changes a user-visible timestamp: `floor(s/3600)` and `floor(floor(s/60)/60)` agree for positive integers, and the boundaries line up (`s < 3600` ⟺ `mins < 60`, `s < 86400` ⟺ `hours < 24`). Two `formatAgo` implementations looking alike is not the same as computing alike — this pair does.

PR #552, opened 2026-08-25
docs: remove stale Solid/tmux/ChatTab references from orchestrator and kobe-daemon

Sma1lboy:
Second-pass comment cleanup for `orchestrator/` + `kobe-daemon/`: remove references to technologies that no longer exist in the codebase (Solid, tmux, ChatTab) so the comments describe the current OpenTUI/hosted-PTY architecture instead of the retired stack.

## What changed
- `orchestrator/core.ts`: "Solid signal" → "Observable state".
- `kobe-daemon` comments updated across:
  - `server.ts` (lifetime policy no longer talks about `tmux attach` / ChatTab windows)
  - `protocol.ts` (v0.6 rationale, PTY host description, `SubscribeRole`)
  - `pty-host.ts`, `pty-server.ts`, `paths.ts` (tmux-analogy wording replaced with "persistent terminal host")
  - `transcript-activity-collector.ts`, `channels.ts`, `collectors.ts` (ChatTab → Terminal Tab; tmux capture-pane → pane-local checks)
  - `handlers.ts`, `handlers-task.ts`, `client-connection.ts`, `auto-title-poller.ts`, `ui-prefs-watcher.ts`, `pty-env.js`

## Verified
- `bun run lint` ✅
- `cd packages/kobe-daemon && bun run typecheck` ✅
- `cd packages/kobe && bun run test:fast` — 348 files, 3431 tests passed ✅

## Left untouched / flagged for human decision
- **`packages/kobe-daemon/src/daemon/server-types.ts`**: a 2-line transitional re-export (`DaemonServer`, `DaemonServerOptions`) with **zero consumers** in any of the three packages. Per the hard rule against deletions, I did not remove it; this PR flags it for Jackson to decide (drop vs. keep as a public compat alias).
- **Cross-area stale comments**: `packages/kobe/src/tui/context/theme/hex.ts` and `packages/kobe/src/tui/lib/apply-ui-prefs.ts` still reference the deleted `Solid store` / `tmux-border-theme.ts`. They are outside this round's scan area (`orchestrator/` + `kobe-daemon/`), so they are reported here rather than edited.
- **`keep in sync` markers** in `ui-prefs-watcher.ts` and `keybindings-watcher.ts` remain — they describe real duplicated constants across package boundaries and are still valid.
- **`handlers-agent-turns.ts` "Socket-only for now"** was checked and still accurately describes the current dashboard behavior, so it stays.


Sma1lboy:
Merging. Two checks before it went in.

**The rebase is clean.** `git diff main..HEAD` is comment-only again, and `open-external.ts` keeps #550's `lib/spawn-detached` import. Worth stating what went wrong the first time, since it is a trap any long-lived branch hits: the three-dot diff (`main...pr`) counts everything on `main` since the merge base, so a stale branch's diff looks like it contains changes it never made. Only the two-dot diff (`main..pr`) answers "what will land if I merge this" — and that one showed this PR silently reverting #550's spawn-detached consolidation. A `docs:` PR whose diff contains code lines is always a signal, never noise.

**The replacements are rewrites, not substitutions.** Spot-checking the daemon comments: `tmux panes` → `OpenTUI terminal panes`, `chattab` → `Terminal Tab`, and "the daemon must never touch tmux" → "never touch front-end state" — that last one is the tell. A find-and-replace would have left the sentence describing a specific dead mechanism; this one generalizes the invariant that outlived it, which is the part still true.

These comments were describing removed subsystems (tmux, Solid) and a renamed concept (ChatTab). That is the harmless half of the pattern this sweep kept hitting — stale *description* rather than a stale *constraint* — but it is still what a new reader learns the architecture from.

PR #555, opened 2026-08-25
chore: add a knip config tuned for this repo's dead-code discovery

Sma1lboy:
A tool, not a gate. Run it when you want to hunt dead code; nothing in CI changes.

## Why a config is the whole point

`npx knip` with no config reports **48 unused files**, nearly all false:

- `packages/kobe-web/e2e/hero-*.ts` — Playwright loads these by path
- `.claude/skills/**`, `.agents/skills/**` — not part of any build graph
- `packages/kobe-landing/*.js` — referenced from HTML `<script>` tags
- `packages/kobe-daemon/src/daemon/pty-host-node-entry.ts` — production reaches it through a **string path** (`const PTY_HOST_NODE_ENTRY = "../daemon/pty-host-node-entry.ts"`), which no static analysis can follow

At that ratio the output is noise. Declaring the real entry points fixes it.

## What the config declares

| workspace | entry |
|---|---|
| `packages/kobe` | both bin wrappers (`rove.ts`, `kobe.ts`), plus the mock hosts `dev:mock` boots |
| `packages/kobe-daemon` | `compat-env`, the whole `client/**` surface, and the node PTY entry |

`kobe-web` is deliberately left out — its Playwright fixtures and HTML-referenced scripts would need entry rules of their own, and that package is a secondary surface.

## Measured result

kobe + kobe-daemon go from 48 candidates to **4**:

| candidate | verdict |
|---|---|
| `daemon/server-types.ts` | **dead** — 2 lines, 0 consumers, self-described "Transitional … retained by the rename branch"; that rename is long done |
| `tui/component/border.ts` | **dead** — `EmptyBorder`/`SplitBorder`/`HSplitBorder` all zero consumers. Its header says it exists "so both the Solid and React pane trees can share it" — Solid was removed 2026-07-07 and React never picked it up (`git log -S"SplitBorder"` shows no consumer ever existed) |
| `cli/launcher.ts` | **false positive** — a build artifact entry; `scripts/build.ts` produces it per bin name, no import reaches it |
| `tui/history/mock-fixtures.ts` | **false positive** — reached through the mock-host chain |

2 real out of 4. Both real ones were confirmed independently before this config existed (issue #46), which is what makes them a fair benchmark rather than a self-fulfilling one.

## Why it isn't a CI gate

The two false positives are exactly the kind a human recognizes in seconds and a rule cannot: "this file is produced by the build" and "this is reached through a mock chain". Gating on that would either block legitimate PRs or push people to write ignore rules until the tool says nothing. Left as something you run deliberately.

Tooling only, empty changeset.

### DIFF
--- a/.changeset/knip-config.md
+++ b/.changeset/knip-config.md
@@ -1,0 +1,4 @@
+---
+---
+
+Tooling only: add a knip config for dead-code discovery. Not wired into CI.
--- a/knip.json
+++ b/knip.json
@@ -1,0 +1,34 @@
+{
+  "workspaces": {
+    "packages/kobe": {
+      "entry": [
+        "src/cli/rove.ts",
+        "src/cli/kobe.ts",
+        "src/**/*mock-host.tsx",
+        "src/tui-react/mock/**"
+      ],
+      "project": [
+        "src/**/*.{ts,tsx}"
+      ]
+    },
+    "packages/kobe-daemon": {
+      "entry": [
+        "src/compat-env.ts",
+        "src/client/**/*.ts",
+        "src/daemon/pty-host-node-entry.ts"
+      ],
+      "project": [
+        "src/**/*.ts"
+      ]
+    }
+  },
+  "ignore": [
+    "**/*.test.ts",
+    "**/*.test.tsx"
+  ],
+  "ignoreDependencies": [
+    ".*"
+  ],
+  "includeEntryExports": false,
+  "$schema": "https://unpkg.com/knip@latest/schema.json"
+}


Sma1lboy:
Closing this — I built something that already exists.

`packages/kobe/knip.json` has been in the repo the whole time, along with a `knip` script (`knip-bun`) and `knip@^6.14.2` in devDependencies. Running the existing setup gives the same answer mine did:

```
$ cd packages/kobe && bun run knip
Unused files (3)
  src/cli/launcher.ts
  src/tui/component/border.ts
  src/tui/history/mock-fixtures.ts
```

Same three candidates, same real finding (`border.ts`). A root-level `knip.json` on top of that is duplicate configuration for one tool — exactly the "check what's already here before writing your own" failure I spent tonight asking agents to avoid. I ran `npx knip` unconfigured, saw noise, and concluded the repo needed a config, without first checking whether it had one.

**The finding survives, the config doesn't.** `tui/component/border.ts` is dead: `EmptyBorder` / `SplitBorder` / `HSplitBorder` all have zero consumers, and its header explains it exists "so both the Solid and React pane trees can share it" — Solid was removed 2026-07-07 and React never picked it up (`git log -S"SplitBorder"` shows no consumer ever existed). That is recorded in the daemon issue store (#46) alongside the other confirmed dead code.

Two things worth leaving behind for whoever runs `bun run knip` next:

- **It exits 1 today** — 133 unused exports, plus `node-pty` and `kobe-web` flagged as unused deps. Some of that is real, some is the tool not seeing string-path and workspace-protocol reachability. It is a report to read, not a gate to pass.
- **`src/cli/launcher.ts` and `src/tui/history/mock-fixtures.ts` are false positives** in that list: the first is a build artifact `scripts/build.ts` produces per bin name, the second is reached through the mock-host chain. Worth knowing before anyone acts on the file list.

Issue #551, opened 2026-08-25
Deleted tasks can resurrect under concurrent peer edits


Sma1lboy:
## 症状

在多实例并发写入 `~/.rove/tasks.json` 的场景下（daemon、CLI、多个 TUI、多个 engine 会话），本进程删除的 task 可能被 peer 的后续编辑“复活”。

## 可复现的测试草图

```ts
// 模拟两个独立的 TaskIndexStore 实例 A 和 B，共用同一个 tasks.json
const a = new TaskIndexStore({ homeDir: tmp })
const b = new TaskIndexStore({ homeDir: tmp })
await a.load()
await b.load()

// A 创建一个 task
const task = await a.createTask({ repo: "/tmp/repo", title: "x" })
await a.save() // 写入磁盘

// B 加载后看到这个 task
await b.load()

// A 删除它
await a.deleteTask(task.id)
await a.save() // disk 上 task 已消失

// B 在不知情的情况下 touch 了该 task（例如改标题、切 vendor、touchRecency）
b.updateTask(task.id, (t) => ({ ...t, title: "edited by B" }))
await b.save() // 会把 task 写回磁盘，导致“复活”

// 预期：task 应该保持删除
// 实际：task 重新出现在 tasks.json
```

## 根因

`TaskIndexStore.removedIds` 是**纯内存**的 `Set`，只在当前实例有效，且 `doSave` 成功后就 `delete` 掉（`store.ts:192`）。`mergeTasksWithDisk` 的逻辑是：

```ts
if (dirty.has(task.id)) {
  result.push(task) // 本实例改过：我们的版本赢
} else {
  const onDisk = diskById.get(task.id)
  if (onDisk === undefined) continue //  untouched AND gone from disk: peer removed it
  result.push(onDisk)
}
```

因此：
- 本实例删除的 task，墓碑只保护到本实例的下一次 save；save 后墓碑消失。
- peer 实例没有该墓碑；如果 peer 把该 task 标成 dirty 并 save，由于其 cache 里还有这个 task，`dirty.has(task.id)` 为 true，会直接把 task 写回磁盘。

### A. 持久墓碑（persistent tombstone）
- 删除时把 id 写入一个持久集合（例如 `tasks.json` 里的 `removedIds` 数组，或单独文件）。
- 合并时任何在墓碑里的 id 都不复活。
- 代价：需要决定何时清理墓碑（任务 id 是 ULID，不会自然过期）；持久集合本身也需要并发保护。

[...]

PR #558, opened 2026-08-26
fix: spell every single weekday correctly in the schedule preview

Sma1lboy:
## Direction

Bugs / correctness — a self-found defect in the Routines/automation schedule preview, in the same "spell the schedule right" class as the earlier weekday fixes (#376, #369).

## Problem

`describeCron` in `packages/kobe/src/tui/component/cron-segments.ts` built a single weekday's plural by capitalizing its three-letter cron abbreviation and appending `"days"`:

```ts
if (/^[A-Z]{3}$/.test(dow)) return `${dow.charAt(0)}${dow.slice(1).toLowerCase()}days ${at}`
```

That only reads correctly for the three days whose full name *is* the abbreviation — `MON`→"Mondays", `FRI`→"Fridays", `SUN`→"Sundays". For the rest it produced mangled words:

| cron day | shown (before) | correct |
|----------|----------------|---------|
| `TUE` | Tue**days** | Tuesdays |
| `WED` | Wed**days** | Wednesdays |
| `THU` | Thu**days** | Thursdays |
| `SAT` | Sat**days** | Saturdays |

Every single-weekday rung is reachable from the schedule picker's day-of-week ladder, and the phrase is rendered live in the automation composer dialog (`automation-composer-dialog.tsx:262`), so a user scheduling a Tuesday routine saw "Tuedays at 09:00". The regex also invented a name for any three-letter code (`XYZ`→"Xyzdays").

## Fix

Map each abbreviation to its full weekday name and pluralize that; return `null` for an unrecognized three-letter code (silence beats a made-up day, matching the function's existing "don't describe a shape I can't model" contract). Net +12 lines of source.

## Verification

- `bun x vitest run test/tui/cron-segments.test.ts` — 15/15 pass. Added a case asserting the correct plural for all seven weekdays (MON–SUN) and one asserting an unknown code (`XYZ`) stays `null`. The prior suite only covered `MON`, which was accidentally correct, so the bug slipped through.
- `bun run lint` and `bun run typecheck` — both green.

## Follow-ups

None for this slice. Unrelated to, and does not touch, the open concurrency-contract discussion in #551.

---
_Generated by [Claude Code](https://claude.ai/code/session_019h3VWAqyBMDyZ8XUTXqV3Z)_

### DIFF
--- a/.changeset/weekday-schedule-preview-spelling.md
+++ b/.changeset/weekday-schedule-preview-spelling.md
@@ -1,0 +1,5 @@
+---
+"@sma1lboy/rove": patch
+---
+
+The schedule preview now spells every weekday correctly — a routine set for Tuesday, Wednesday, Thursday, or Saturday reads "Tuesdays"/"Wednesdays"/"Thursdays"/"Saturdays" instead of the mangled "Tuedays"/"Weddays"/"Thudays"/"Satdays".
--- a/packages/kobe/src/tui/component/cron-segments.ts
+++ b/packages/kobe/src/tui/component/cron-segments.ts
@@ -41,6 +41,17 @@ function range(from: number, to: number): string[] {
   return out
 }
 
+/** Full weekday name for a cron three-letter abbreviation (uppercase). */
+const WEEKDAY_NAMES: Record<string, string> = {
+  MON: "Monday",
+  TUE: "Tuesday",
+  WED: "Wednesday",
+  THU: "Thursday",
+  FRI: "Friday",
+  SAT: "Saturday",
+  SUN: "Sunday",
+}
+
 const LADDERS: Record<CronSegment, readonly string[]> = {
   minute: MINUTE_LADDER,
   hour: HOUR_LADDER,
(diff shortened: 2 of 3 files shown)


Sma1lboy:
Closing this as a duplicate. After opening it I found that the exact same `describeCron` weekday-spelling fix is already covered by two existing open PRs — #376 ("spell every weekday correctly in the schedule preview") and #369 ("spell every weekday correctly in the Routines schedule preview"). All three replace the `${dow.charAt(0)}${dow.slice(1).toLowerCase()}days` construction with a weekday→name map and add the same TUE/WED/THU/SAT tests, so this PR adds nothing beyond them.

#376 is the most complete of the three — it also fixes `describeTimeOfDay` staying silent on a list/range minute (`15,45 * * * *`) rather than asserting a `:15,45` fire time — so that's the one worth reviewing. Sorry for the noise; my pre-flight survey didn't connect those PR titles to this code before I started.

---
_Generated by [Claude Code](https://claude.ai/code)_

PR #568, opened 2026-08-27
fix(store): atomic token lockfile, per-save tmp, persistent deletion tombstones

Sma1lboy:
Fixes rove issues #53 and #47 — two faces of the same shared write path (`~/.rove/tasks.json`), smaller together than apart.

**Not for merging tonight** — parked for review; see the stack note at the bottom.

## #53 — the lock did not actually exclude

The chosen direction is *tighten the lock*, not *tmp-only*, because the real mutex break is upstream of the tmp collision: `writeFile(lockPath, pid, {flag:"wx"})` **creates the lockfile before its content lands**. A reader in that window sees an empty file, parses pid `-1`, classifies a live holder as stale, and steals the lock. From there the vanished-lock retry and the shared tmp turn into the CI flake's `rename tasks.json.tmp ENOENT`.

- Lock contents are written to a private sidecar and `link(2)`ed into place — never observable half-written.
- Contents are `pid:token`, so two stores in the **same process** are distinct holders (`parseInt` still reads the pid, so old builds and legacy bare-pid locks stay compatible in both directions).
- `release` only unlinks a lock it still owns, so a forced takeover cannot cascade a third writer into the critical section.
- Each save also stages through a **unique tmp file** — the issue's other direction, kept as defense: a peer running an old build with the broken lock can no longer make our `rename` crash.

## #47 — deletion vs concurrent edit

Chose persistent tombstones over documenting last-write-wins. The racing "edit" is nearly always a trivial recency/status bump; the delete is a deliberate user action that already destroyed the worktree, so a resurrected row is a ghost the user has to delete again.

Deletions persist as an optional `removed: [{id, at}]` field that beats a concurrent dirty edit **from either side**, pruned after 30 days (a tombstone only has to outlive live writers' in-memory dirty refs). Task ids are ULIDs and never reused, so a tombstone cannot shadow a legitimate new task.

`version` stays **3** on purpose: older builds ignore the field on read and drop it on write, degrading to today's behavior — whereas a version bump would make them treat the whole manifest as unsupported and recover empty.

## Tests — deterministic, not stress

The original flake needed a full 8-worker suite to reproduce once, so neither test relies on winning a race:

- `store-tmp-collision.test.ts` pauses one save between its tmp write and its rename (fs mock), force-breaks the lock the way the old steal did, completes a rival save, and reproduces the exact `ENOENT` against the old shared-tmp code.
- `store-concurrency.test.ts` uses `touchRecency` to hold a deleted task dirty across a peer's `remove`.

Both verified red against the reverted fix. Full `test:fast` (3484 tests), lint, and typecheck green.

---

Stacked context: opened alongside #567 (`renameBranch` convergence). Independent file sets, no conflict — both are parked for review rather than merge.

### DIFF
--- a/.changeset/task-index-lock-tombstones.md
+++ b/.changeset/task-index-lock-tombstones.md
@@ -1,0 +1,5 @@
+---
+"@sma1lboy/rove": patch
+---
+
+Fix two concurrency holes in the shared task-index write path. The index lockfile is now created atomically (content-first sidecar + link) and carries a per-acquire token, so a reader can never observe a half-written lock and classify a live holder as stale, two stores in the same process are distinguishable holders, and release can no longer unlink a lock a rival legitimately took over — closing the mutual-exclusion break behind the `rename tasks.json.tmp ENOENT` CI flake (#53); each save also stages through a unique tmp file so writers can't clobber each other's staging even under a broken mutex. Deletions are now persisted as tombstones in `tasks.json` (optional `removed` field, pruned after 30 days), so a peer instance that still holds a deleted task dirty in memory can no longer resurrect it on its next save (#47).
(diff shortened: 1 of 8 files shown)


Sma1lboy:
This also fixes #551 (the public report of the resurrection half — rove issue #47).

PR #590, opened 2026-08-27
fix(plugins): stat-poll the registry so a write during watcher startup is never lost

Sma1lboy:
Root-causes and fixes the largest contributor to rove issue #61 — the intermittent `test:fast` failures I hit repeatedly tonight.

**Not for merging tonight** — parked for review.

## It was a real bug, not test flakiness

On macOS the FSEvents stream behind `fs.watch` **starts asynchronously**. A `plugins.json` write landing in that startup window is **dropped forever, with no signal** — the watcher simply never fires.

Instrumentation showed `fs.watch` delivering **zero events in every failing run**. The falsifiable check: inserting a 250 ms delay before the write took 96 stressed runs from 7–9 failures to **0**. That confirms the window, rather than just correlating with it.

This matters beyond tests: **a `rove plugin install/link/enable` racing daemon startup would be silently ignored** until the next restart.

## The fix is correct by construction, not by widening a window

A synchronous baseline `stat` is taken **before** the first registry load, then polled every 200 ms:

- a write landing *before* the baseline is seen by the load itself
- a write landing *after* it flips the stamp and triggers a reload

No write can fall between the two. That's the property `fs.watch` never had here.

Per the brief this deliberately avoids "add a sleep / raise the timeout / retry" — the same discipline that turned PR #568's "CI flake" into a genuine concurrency defect.

## Verification

I ran my own before/after rather than relying on the report:

- **24 rounds of the sample test**: unfixed → 1 failure; fixed → **0**.
- **8 full `test:fast` runs on the fixed branch**: 7 fully green, 1 showing 2 failed.

**So this is a real improvement but not a total cure.** Before the fix I was seeing a failure roughly every third full run tonight; now it's roughly 1 in 8, and the residual failures did not reproduce often enough in this session to name. Issue #61 should stay open for the remainder rather than be closed on this PR.

## Related, deliberately untouched

The worker flags that `file-watch-trigger.ts` (issue #3's flake) and the filetree pane use the **same `fs.watch` pattern** and are therefore in the same bug class. Not changed here — that's a separate scope call.

`test:fast` green on repeat runs, lint and typecheck clean. `docs/design/plugins.md` updated in the same change.

### DIFF
--- a/.changeset/plugin-registry-poll-watch.md
+++ b/.changeset/plugin-registry-poll-watch.md
@@ -1,0 +1,5 @@
+---
+"@sma1lboy/rove": patch
+---
+
+Watch the plugin registry by stat-polling instead of `fs.watch`. On macOS the FSEvents stream behind `fs.watch` starts asynchronously, so a `plugins.json` write landing before the stream is live was dropped forever — the daemon never saw the install/enable until the next mutation, and the plugin-runtime test flaked ~8% under a loaded suite (issue #61). The host now takes a synchronous baseline stat before the first registry load and polls every 200ms, so no write can fall between the watcher and the load.
--- a/docs/design/plugins.md
+++ b/docs/design/plugins.md
@@ -106,8 +106,10 @@ local authoring (no build; your tree, your build). `uninstall` removes the
 managed checkout but keeps `config/` + `state/`; `unlink` never touches files.
 
 Registry: `~/.kobe/plugins.json`, written only by the CLI. The daemon
-(`plugins/runtime.ts`, wired in `daemon/server.ts`) file-watches it, so
-install/enable/disable apply to a running daemon without a restart. Run log:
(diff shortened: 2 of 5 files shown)


Sma1lboy:
**CI note — this run's `typecheck-and-test` failure is issue #61's residue, not this change.**

The failing test is `terminal-pty-scrollback-cache.test.ts > publishes a stable absolute window origin across saturated scrollback shifts` — the terminal PTY scrollback cache, which shares no code with the plugin registry. Checked rather than assumed:

- The test file contains **no reference to plugins**, and this PR touches **no terminal/scrollback file**.
- Run **12 times in isolation on this branch: 0 failures.** It only fails under the full concurrent suite — the exact signature described in #61.

I'm deliberately not re-running until it goes green and calling that a pass. Reflexively re-running red is the failure mode #61 is about, and this PR only removes the *largest* contributor (macOS `fs.watch` startup window), not all of them — which is why #61 stays open with the residual measured at roughly 1 full run in 8.

PR #594, opened 2026-08-27
perf(landing): lazy WebP final band, defer scripts; a11y contrast + landmarks

Sma1lboy:
A performance and accessibility pass over the landing page. Deliberately scoped **away** from copy, hero structure, and layout — a parallel change is working that area, and this one doesn't touch it.

**Not for merging tonight** — parked for review.

## Performance

The final band's background was a **478 KB PNG loaded via CSS `background: url()`** — render-blocking, fetched on every first paint, for decoration below the fold. It's now a **39.6 KB WebP** in an ``. Scripts that don't need to parse before first paint are deferred.

**I measured this myself rather than repeating the report.** My first attempt found *no* difference, because my script only counted `` tags and the old asset was a CSS background — worth knowing, since that's exactly the kind of load an audit misses:

| | first-paint bytes |
|---|---|
| before | 82 KB (css+js) + **467 KB** blocking background = **549 KB** |
| after | 82 KB (css+js), image lazy = **82 KB** |

WebP is **92% smaller** than the PNG it replaces. (The report cites 625→148 KB on a wider basket; the direction and magnitude hold either way.)

## Accessibility

Contrast: `accent`, `faint`, `ok`, `warn` had their **lightness lowered only** — hue and chroma untouched, so this is tuning inside the brand, not a reskin. Verified by computing the ratios rather than trusting the claim:

- old `oklch(62% 0.13 45)` on paper → **3.59:1 — fails AA**
- new `oklch(56% 0.13 45)` on paper → **4.60:1 — passes AA**

Also added: `<nav aria-label="Main">` and an Inbox landmark, `role="dialog"` + `aria-modal` on the modal, `aria-live="polite"` for dynamic status, and `aria-hidden` on decorative nodes. `prefers-reduced-motion` blocks are present across the animated stylesheets.

## Checks

- **i18n parity intact** — zero one-sided keys between the zh/en dicts.
- Scope confirmed: CSS/HTML/asset changes only; no hero restructuring, so it shouldn't conflict with the parallel copy/mobile pass.
- Lint clean.


Sma1lboy:
**CI fix pushed.** `coverage-cap` / `typecheck-and-test` were failing on a real assertion, not a flake:

`test/architecture/package-distribution.test.ts > landing pages load their extracted static assets` pinned the **literal** tag `<script src="/index.js"></script>`, so adding `defer` — the whole point of this change — broke it.

The test's intent is *"the page references its extracted asset"*; the exact spelling was incidental. It now matches the `src` reference instead. **Verified it still bites**: deleting the script tag entirely makes it fail (1 failed / 18 passed), so this relaxes the assertion to its intent without weakening the guard.

PR #608, opened 2026-08-28
fix(doctor+capture): finish the hook-channel check and re-shoot the plugin clips

Sma1lboy:
Six commits that accumulated on this branch after #565 (`fix(doctor): report when the engine hook channel is dead`) was merged from its first commit. They are the follow-through on that check plus the plugin-clip re-shoot, and they carry the two changesets that never reached main.

## Doctor — finishing the hook-channel check

#565 landed the check; these four commits finish it:

- **`fix(doctor)`: say so when the hook check itself cannot run.** A check that cannot read the registry was indistinguishable from a healthy channel — it now reports that fourth state instead of implying "live".
- **`fix(doctor)`: drop the unobservable socket-override check.** It claimed to detect a stale per-tab socket override, but nothing in the recorded data can distinguish that from a quiet channel. A check that cannot fail is worse than no check.
- **`chore(doctor)`: drop the unread `state` field from the inspect tab shape.**
- **`docs(changeset)`: describe the check that shipped, not the one that was cut.** The changeset still advertised the removed socket-override probe.

## Plugin example clips

- **`docs(plugins)`: re-shoot the example clips in the real TUI.** The old GIFs filmed a shell running `rove api …` and `cat`-ing a log — true, but no frame contained the product, and the two examples with the most to show (a pane, a contributed engine) had no clip at all. Each example is now recorded where its surface actually appears. Captures ride the same browser-PTY path as the README assets.
- **`fix(capture)`: render recordings through WebGL so block glyphs tile.** Canvas rendering left seams between block glyphs in the captured frames.

Two guards came along with the re-shoot, both of which protect the operator rather than the output: the hero fixture's daemon socket is now pinned rather than derived (a compatibility symlink under `.kobe/` could otherwise point a capture at the live daemon), and a take aborts if the TUI ever displays an e-mail address — Settings → Engines renders real accounts, since `HOME` stays the operator's by design.

## Changesets

Two, both `patch`: `doctor-hook-channel-probe` and `plugin-demo-clips`.

## Verification

`bun --filter @sma1lboy/rove test:fast` — 3479 passed. `bun run lint` and `typecheck` clean. Re-verified against current main by CI here, since main moved ~40 commits while this branch sat.


Sma1lboy:
Superseded by #610. This branch's baseline had rotted ~50 commits behind main, and its four `doctor` commits had already reached main through #565 in their final form (rebasing them conflicted against work that no longer exists there). #610 carries the two commits that are genuinely new — the plugin clip re-shoot and the WebGL capture fix — cherry-picked onto current main.

PR #624, opened 2026-08-29
fix: parse numstat renames with --numstat -z

Sma1lboy:
Closes #63.

Switch the three file-tree `git diff --numstat` callers to `--numstat -z` and update the shared parser to consume NUL-delimited records. This removes the brace-compaction parser, which is inherently ambiguous when a path contains a literal `{`.

## Before / after for the #63 counterexample

Path rename: `src/{a{b/f.txt` → `src/c/f.txt`

```text
# git diff --numstat (non -z)
0\t0\tsrc/{{a{b => c}/f.txt
# Three `{` — no anchor position can reliably distinguish the brace syntax
# from the literal brace in the path.

# git diff --numstat -z
0\t0\t\\0src/{a{b/f.txt\\0src/c/f.txt\\0
# Old and new paths are independent NUL-separated fields.
```

## Callers updated

- `src/tui/panes/filetree/git.ts:142` — `diff --numstat -z HEAD`
- `src/tui/panes/filetree/git.ts:152` — `diff --numstat -z --cached`
- `src/tui/panes/filetree/git.ts:284` — `diff --numstat -z <base>...HEAD`

The untracked-file merge logic at `filetree/git.ts:175` is unaffected: it joins numstat counts onto porcelain rows by canonical path, and `-z` emits the same canonical paths.

## Verification

- Added regression test for the literal-brace rename path.
- `bun run typecheck`, `bun run lint`, `bun run test`, and `bun test test/render` all pass.

Note: this PR stays open for review; I will not merge it myself.

### DIFF
--- a/.changeset/numstat-z-rename.md
+++ b/.changeset/numstat-z-rename.md
@@ -1,0 +1,13 @@
+---
+"@sma1lboy/rove": patch
+---
+
+Fix numstat rename parsing for paths containing literal braces
+
+Switch `git diff --numstat` callers to `--numstat -z` so renames are emitted
+as NUL-delimited old/new path pairs instead of brace-compacted `src/{old =>
+new}` syntax. The brace form is inherently ambiguous when a path itself
+contains `{` (e.g. `src/{a{b/f.txt`), which previously produced silently
+wrong filenames. All three file-tree numstat consumers (working diff, cached
+fallback, and branch-scope diff) now pass `-z`; the shared parser in
+`src/lib/git-parsers.ts` no longer tries to split brace-compacted fields.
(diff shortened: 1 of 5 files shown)


Sma1lboy:
Update: fixed the render-track pollution reported above.

Root cause: `test/render/host-pages.test.tsx` rendered `SettingsDialog` / `WorktreesPage` inside its `RenderHarness` without `DialogProvider`, so `useDialog()` threw. The ErrorBoundary usually caught it, but the thrown render leaked state across render tests and flakily broke the later `worktrees-page-delete.test.tsx` (`a failed delete puts the row back`).

Fix: add `providers: { dialog: true }` to the two `RenderHarness` tests that mount dialog-consuming pages.

Verification:
- `host-pages.test.tsx` + `worktrees-page-delete.test.tsx` together: 10/10 green.
- `bun test test/render`: green in latest run.

Note: while stress-testing I also hit an unrelated flake in `new-chat-flow.test.tsx` (`fork+continue leads the child's first prompt with the handoff brief`, ~1 in 10). It is unrelated to numstat / git-parsers; I am leaving it for a separate issue.

PR #636, opened 2026-08-30
fix: let each engine own its session id and resume verb

Sma1lboy:
94% of engine tabs are kimi, and none of them survived a restart with their conversation. On disk: of 366 engine tabs, 7 had a `sessionId` — all claude.

## The two breaks

**No id was ever recorded.** `withClaudeSessionId()` opened with `coerceVendorId(vendor) !== "claude"`, so only an engine *literally named* claude got a `--session-id` pin. Kimi got none. Neither did wrapper presets like `claudecpa` — a zsh function around the real claude, excluded by its name rather than by what it runs.

**The resume flag was hard-coded.** `engineTabArgv` appended `--resume` unconditionally. Kimi's flag is `-S`; even with an id, the launch would have died.

## The shape

Each engine now declares its own session verbs (`engine/session-identity.ts`). Neutral layers ask the engine instead of naming a vendor — the old helper's *name* was itself the AGENTS.md violation.

An id has three possible origins, and which one applies is a property of the CLI, not a choice Rove makes:

| origin | engine | mechanism |
|---|---|---|
| pinned at launch | claude | `--session-id <uuid>`, resumes `--resume <id>` |
| read from OSC title | codex | existing `sessionIdFromTitle`; resumes via the `resume` subcommand |
| discovered after the fact | kimi | store lookup by worktree; resumes `-S <id>` |

Kimi's CLI *cannot* be told what to call a new session — `-S [id]` only reopens an existing one. So its id is discovered from the session store the history reader already indexes. A custom preset inherits the verbs of the protocol it declares, which is what fixes `claudecpa`. An engine with no resume verb starts a fresh conversation honestly rather than being handed a flag that would kill its launch.

Discovery is claim-tracked: the store answers per-*worktree*, so without it two kimi tabs in one task would both adopt the newest session and fight over one conversation.

Also fixes the restart existence check, which asked `readHistory(id).length > 0` — that silently means "Rove can parse this engine's messages". Kimi ships no parser, so every kimi session reported as absent and its tab respawned blank even once its id was known.

## Verification

Two kimi engines were killed by mistake while verifying this change; their conversations are recoverable only through the `kimi -S` path this PR adds.

Claude and codex were pinned with characterization tests written *before* the change, so their shapes are held against regression rather than re-derived afterwards.

Probed against the real binaries (2026-08-29), in an isolated sandbox:

- `kimi -S <id>` resumed and recalled the earlier turn; a bare launch replied *"this is the first message in our conversation"* — the bug and the fix, side by side.
- A real kimi tab's id was discovered and recorded (`session_fb636b17…`, `spawned: true`) and survived a daemon restart.

Testing surfaced a real wiring bug: discovery first sat in the 5s naming poll behind a `lastTitle` gate that kimi never sets (alt-screen, no OSC title), so it ran *after* the tab had already respawned blank. It now runs in the hydration gate, which is what holds the spawn back.

`registry.ts` went over the 500-line cap, so its built-in table moved to `builtin-engines.ts` — same pattern as `history-readers.ts`. Every touched file is now under the cap.

Suites: 209 engine/tui vitest files (1943 tests) and 235 render tests pass; typecheck and lint clean.

coverage-exemption: packages/kobe/src/tui-react/workspace/TerminalTabs.tsx — integration layer needing a live daemon + PTY; the render track cannot mount it. Covered by the behavior track and the visual-ground-truth harness. Pre-existing gap (8.87% on main), and this PR only changes one import line and one call.
coverage-exemption: packages/kobe/src/tui/workspace/terminal-tab-argv.ts — pure argv composition, covered by vitest (test/tui/terminal-tab-fork-argv.test.ts and the new test/tui/engine-tab-resume-argv.test.ts), which the render-track lcov does not see. Pre-existing gap (6.58% on main), not introduced by this PR.

### DIFF
--- a/.changeset/engine-owned-session-identity.md
+++ b/.changeset/engine-owned-session-identity.md
@@ -1,0 +1,9 @@
+---
+"@sma1lboy/rove": patch
+---
+
+Every engine now answers "what is my session id" and "how do I resume it" for itself, so a restart reconnects a kimi or wrapper-engine tab to its conversation instead of opening a blank one.
+
+Session identity was hard-coded to Claude in two places: the launch pinned an id only when the vendor was literally named `claude`, and the restart path always appended Claude's `--resume` flag. Kimi tabs therefore never recorded an id at all — and neither did custom wrapper engines like a `claudecpa` preset, which is a Claude launch under another name. Each engine now declares its own session verbs (`engine/session-identity.ts`): Claude pins `--session-id` and resumes with `--resume`, Codex resumes through its `resume` subcommand, and Kimi — whose CLI can only reopen an existing session, never name a new one — has its id discovered from its session store after the fact and resumes with `-S`. A custom preset inherits the verbs of the protocol it declares. An engine with no resume verb starts a fresh conversation honestly rather than being passed a flag that would kill its launch.
+
+Also fixes the restart existence check, which asked whether Rove could parse a session's messages rather than whether the session existed — so engines that ship no message parser reported every conversation as absent.
--- a/packages/kobe/src/cli/api/runtime.ts
+++ b/packages/kobe/src/cli/api/runtime.ts
@@ -7,8 +7,8 @@
  */
 
 import type { PtySessionExit } from "@sma1lboy/kobe-daemon/daemon/protocol"
-import { engineLaunchArgv } from "../../engine/engine-presets.ts"
-import { withClaudeSessionId } from "../../engine/interactive-command.ts"
+import { engineLaunchArgv, withPinnedSessionId } from "../../engine/engine-presets.ts"
+
 import { buildEngineSessionLaunch } from "../../engine/session-launch.ts"
 import { trustEngineWorktree } from "../../engine/trust-worktree.ts"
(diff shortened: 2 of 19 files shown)


Sma1lboy:
CI: 6/7 green. `visual-ground-truth` fails identically on `main` — same 3 specs (`sandbox-keyboard-hints`, `sandbox-workspace-help`, `sandbox-worktree-audit`), same 5 `toContainText` failures, same error class. Verified against main's run 33282925029; pre-existing, not introduced here.

`behavior` and `render-track` both failed on the first run and are now green: the first was a stale base (rebased onto 14ec99d2, which touches the `api send` path that test exercises), the second was a real coverage regression in `use-tab-lifecycle.ts` (48.9% → 90.9% via a render test that mounts the hook against a real on-disk kimi store).

Sma1lboy:
Correction to my earlier CI comment: the failing specs are **5**, not 3 — `e2e/sandbox.spec.ts` lines 80, 104, 119, 141, 167. Identical set on `main` (run 33282925029). My "3" counted deduplicated screenshot-artifact name prefixes, which truncate and collide across tests, not test cases. Conclusion is unchanged: pre-existing, same 5 specs, same error class.

PR #685, opened 2026-08-30
fix: show engine deaths, and tell a rate limit apart from a crash

Sma1lboy:
Rove recorded almost everything about a dying engine and rendered almost none
of it. Four gaps, all display-and-channel rather than detection.

## 1. Engine death was a UI no-op

A killed engine (SIGTERM: no Stop, no SessionEnd, no hook of any kind) reached
the UI as `applyRest(… "no engine in foreground")` — folded into idle. A dead
tab rendered **byte-identically to a shell that had never run anything**. The
exit record (code, signal, the 403 text) was already on disk in
`pty-exits.json`; its only consumer was `rove api inspect`, and the
`session.exited` plugin event had no subscribers outside the plugin host.

The record now travels: `pty-exit-watch` → `DaemonActivityRegistry`
(`recordEngineDeath`) → a new `dead` activity state → `†` in the sidebar and
the tab strip, plus its own durable Inbox episode. The state carries the exit
code and the last error line, so the death says *why*.

The observer keeps folding a vanished foreground into idle, deliberately: the
shell outlives its engine, and quitting your agent on purpose is an ordinary
idle tab. `dead` comes only from the exit record, which knows the difference.

## 2. `rate_limited` and `error` were the same `!`

`turn-state-merge.ts` collapsed both into `error`, while the sidebar has always
drawn them apart (`◷` vs `×`) — two surfaces, one tab, two answers. "Blocked
until the window resets" and "broke, needs you" ask for opposite actions. The
strip now uses the rail's own glyphs.

`Task.quotaResume.resumeAt` was persisted by the daemon's quota probe and read
by **nothing** (zero hits across the TUI and web). A rate-limited Inbox card
now shows `resumes 3:14 PM`, so a wait is distinguishable from a dead end.

## 3. The sidebar computed explanatory words and discarded them

`activityLabelFor` produced "rate limited" / "needs permission" / "error" for
`subtitleText`. The one-line tree row renders only the glyph and the tab label
— the words had maintenance cost and no reader, asserted solely by tests.

Deleted the text; **kept the tone**, which is not dead: it is what keeps a
row's warning/error colour when something else makes it spin (`loading` would
otherwise paint it `primary`). That is why the glyphs must stay distinct —
with no subtitle there is no second channel to disambiguate them.

## 4. Kimi could never classify a failure as a rate limit

`kimi-local` returned no detail for `turn-failed`, so `reduceActivity` saw
`failure: undefined` and every Kimi failure — including the 5-hour wall that
triggered this work — reduced to a generic `error`, never reaching
`rate_limited` and so never arming auto-resume. Now classified, against Kimi's
real vocabulary (verified in the installed 0.37.2 binary): `error_type` is a JS
class name, and the `[provider.*]` code rides in the message. The 403 quota
wall maps to `billing` — attention, but deliberately no resume timer.

**Codex is left alone, and that is the finding.** Its hook enum (verified in
codex-cli 0.149.1) has no failure event at all — `TurnFailed` exists but is an
app-server thread event, not a hook. A classifier there would be dead code. Its
quota probe is the only path. **Kimi still has no quota probe** — out of scope
here, and still missing.

## Verification

- Three runners green: 3886 vitest / 639 socket / render.
- **Nine mutations, each → red**, hitting wires as well as leaves: dropping
  `activity` from the server's watch options; deleting the `publishDeath` call;
  reverting the merge fold; reverting `TURN_GLYPHS` (a *different* site for the
  same slice); dropping Kimi's `turn-failed` branch; removing one classifier
  branch; stopping the `quotaResume` read; deleting the pane's call site;
  dropping the Inbox death write. Re-verified after rebase.
- Two of those exist because a behavioural test could not see a disconnect —
  the watcher tests pass their own `activity`, so `server.ts` dropping it stays
  green. Those are asserted against the wiring itself.
- Screenshots below, via `/harness` → xterm.js → PTY sidecar → real OpenTUI,
  driving a **real Claude Code engine** killed by its exact pid from
  `rove api inspect`.

Note for reviewers: full-suite runs are flaky under parallel load on this
machine — clean `origin/main` failed a different test in the same comparison.
Every file involved passes in isolation.

### DIFF
--- a/.changeset/engine-death-visible.md
+++ b/.changeset/engine-death-visible.md
@@ -1,0 +1,20 @@
+---
+"@sma1lboy/rove": patch
+---
+
+Show engine deaths, and tell a rate limit apart from a crash
+
+A killed engine used to be a UI no-op. The exit record (code, signal, the
+403/quota text) was already written to `pty-exits.json`, but the only consumer
+was the CLI: the observer folded the death into idle, so a dead tab rendered
+identically to a shell that had never run anything. Deaths now reach the
+sidebar and the tab strip as `†`, and land in the Inbox as their own episode.
+
+The tab strip drew `rate_limited` and `error` with the same `!`, while the
+sidebar has always drawn them apart — one tab, two surfaces, two answers. The
+strip now uses the rail's own glyphs (`◷` rate limited, `†` dead), and a
+rate-limited Inbox card shows when its auto-resume fires ("resumes 3:14 PM")
+— a time the daemon has always persisted and nothing ever displayed.
+
+Kimi can now classify its own failures, so a Kimi rate limit reaches
+`rate_limited` (and arms auto-resume) instead of reducing to a generic error.
--- a/docs/TUI.md
+++ b/docs/TUI.md
@@ -56,8 +56,9 @@ on the **tab rows** underneath:
 | `?` | Needs your input: a permission prompt or a question |
 | `●` | Turn finished, and you haven't looked yet |
 | `○` | Idle or not yet observed. Includes a finished turn you've already seen |
-| `◷` | Rate limited |
+| `◷` | Rate limited — the engine is waiting out a quota window |
 | `×` | Error, including a failed worktree deletion |
+| `†` | The engine process exited or was killed |
 | `·` | Not an agent tab, or a custom engine without activity tracking |
 
 A tab labelled `⚠ <name>` is a live hosted session that was missing from the
@@ -73,9 +74,14 @@ completion has a later timestamp and appears unread as usual.
 
 Each tab row reports its **own** activity, not the task's roll-up. Tab 2 can
 spin while tab 1 rests. The tab strip at the top of the workspace uses a
-similar vocabulary (`●` running, `✓` done, `!` error, `?` needs input, `○`
-idle) over the same saved timestamps: a `✓` you have already read settles
-back to `○`, and stays settled after a restart.
+same vocabulary (`●` running, `✓` done, `!` error, `◷` rate limited, `†`
+exited, `?` needs input, `○` idle) over the same saved timestamps: a `✓` you
+have already read settles back to `○`, and stays settled after a restart.
+
+`†` and `×` are different questions. `×` is an engine that ran and reported a
+failed turn — read the tab. `†` is an engine that is no longer running, so the
+answer is to start it again; the exit code and the last error line are kept
+(`rove api inspect`) even after the session is gone.
 
 ## Managing Tasks in the sidebar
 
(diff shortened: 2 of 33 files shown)


Sma1lboy:
## Screenshots — `/harness` → xterm.js → PTY sidecar → real OpenTUI

A **real Claude Code engine** in the fixture task, killed by its exact pid from
`rove api inspect` (`claude --session-id …`, pid 19529). No mocks.

Image files are on the machine that ran the capture (`gh` cannot attach images
to a comment; only the web UI can). Paths:
`/tmp/BEFORE-dead.png`, `/tmp/AFTER-dead.png`, `/tmp/AFTER-ratelimit.png`.
The captured buffer text is quoted verbatim below, which is the part that
carries the evidence.

### 1. Engine death — before

Sidebar row and tab strip, from the captured OpenTUI buffer:

```
▌ · claude 1            │ │ ○ claude 1 │
```

`·` is `NO_STATE_GLYPH` — the same dim dot a shell tab that never ran anything
wears. Nothing in the UI says an agent died here.

### 1. Engine death — after

```
▌ † claude 1            │ │ † claude 1 │
```

`†` in error red, in **both** the sidebar rail and the tab strip. The terminal
pane below reads `⚠ Engine exited (code 143). Check Settings → Engines and fix
the launch command.` — the 2026-08-30 signature.

Daemon state through the real chain (exit record on disk → watcher → registry):

```json
"tab-1": { "state": "dead", "vendor": "claude",
           "exit": { "code": 143, "signal": null, "lastLine": "…" },
           "source": "hook" }
```

### 2. Rate limited

```
▌ ◷ claude 1            │ │ ◷ claude 1 │
```

`◷` in warning amber, in **both** surfaces — the strip previously drew this as
`!`, identical to an error. Driven by a real `rove hook turn-failed --engine
claude` with `error_type: rate_limit`; the daemon reduced it to
`"state": "rate_limited", "source": "hook"`.

## What I could NOT photograph, and why

**The Inbox card with `resumes 3:14 PM`.** The visual fixture's Inbox stayed
empty: `handlers-engine-report` only records an episode when `explicitId` is
set, and I could not get the fixture's hook invocation to satisfy that gate
from outside a real engine process. Rather than fake it, the behaviour is
asserted directly instead:

- `test/tui/inbox-item-view.test.ts` — the note renders as a formatted clock,
  is absent when no resume is armed or the state is not a rate limit, and shows
  nothing (not `Invalid Date`) for a garbage stamp.
- `test/render/tab-strip-state-glyphs.test.ts` — the pane actually calls it and
  puts it in the rendered subtitle (deleting the call site turns this red).
- `test/daemon/inbox-death-episode.test.ts` — the death episode persists with
  its exit detail.

**A death record produced by a real kill.** Killing the engine inside a live
PTY writes *no* exit record — the PTY (the wrapping shell) survives, exactly
the shape of the original incident. And the harness's node-pty sidecar is a
different layer from the production pty-host that writes `pty-exits.json`, so
it cannot author one either. The record behind the "after" capture was written
to disk directly; **everything downstream of it is the real code path** — the
daemon's own watcher parsed it, the registry arbitrated it, the TUI rendered
it. That the engine-inside-a-live-PTY case leaves no record at all is a real
gap, and it is not one this PR closes.

PR #742, opened 2026-09-01
feat: close terminal tabs through the api

NarwhalChen:
## Summary

Adds a `rove api tab-close --task-id <id> --tab tab-N` verb so a Terminal Tab can be closed from the CLI instead of only via ctrl+w in the TUI.

- **Attached TUI**: the daemon brokers the request to the attached client (`terminalTab.close` → TUI's existing ctrl+w close path → `closeReply`), so close semantics — including last-tab behavior — are exactly the manual ones.
- **Headless**: reuses the same `closeTab` pure transition on the persisted tab snapshot (fresh-state transaction so a stale CLI snapshot can't clobber a newer TUI tab list), then kills every hosted PTY session the tab owns via the PTY host.
- Unknown/already-gone tabs fail loud with `TAB_NOT_FOUND` + hint + `nextCommandArgs`.
- Registered in `rove api schema`; docs (`docs/API.md`), skill reference (`references/api-flags.md`) and skill version bumped in the same change; patch changeset included.

## Verification

Local, in the task worktree: `bun run lint` ✓, `tsc --noEmit` ✓, vitest cli suites (100 tests) ✓, `KOBE_INCLUDE_SOCKET=1` daemon handlers (28) ✓, bun-test render suite (10) ✓.

### DIFF
--- a/.agents/skills/kobe/SKILL.md
+++ b/.agents/skills/kobe/SKILL.md
@@ -3,7 +3,7 @@ name: rove
 description: Use when controlling Rove tasks, parallel coding attempts, hosted agent sessions, task lifecycle, or the daemon-owned issue tracker from a shell. Also the ONLY channel for messaging another agent session on this machine — `rove api send`, never a peer/MCP side channel.
 ---
 
-<!-- rove-skill-version: 38 — bump in lockstep with KOBE_SKILL_VERSION (src/lib/skill-install.ts). -->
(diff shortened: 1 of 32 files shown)


Sma1lboy:
Rebased onto current main and reopened as #767 (branch `feat/api-tab-close-verb`); this PR is superseded.

PR #743, opened 2026-09-01
feat: reveal direct shortcuts on ctrl hold

NarwhalChen:
## Summary

Hold a bare Ctrl key (left or right) for 400ms and the direct-shortcut guide opens — the long-press counterpart of the `ctrl+a` prefix board. Release, or any other key, closes it. A quick tap does nothing.

- Requests kitty keyboard protocol flags `events` + `allKeysAsEscapes` (`useKittyKeyboard`, host-render-options), so kitty-protocol terminals (iTerm2 3.5+, kitty, Ghostty, WezTerm) report bare modifier press/release. Terminals without the protocol ignore the negotiation and behave exactly as before — the feature silently degrades (Terminal.app, inside tmux).
- Hold detection is a standalone, unit-tested state machine (`src/tui/lib/ctrl-hold.ts`, threshold constant 400ms).
- Kitty modifier keycode names (57441–57454 family, `src/tui/lib/modifier-keys.ts`) are filtered at dispatch/passthrough so they never leak into the embedded PTY or composer.
- The guide reuses the prefix-HUD overlay and derives every currently reachable direct binding from the live `KobeKeymap`. That includes high-frequency pane keys such as `n`, `q`, `j`/`k`, `r`, and `d` in the sidebar, without maintaining a second hardcoded shortcut list.
- Overlay height accounts for wrapping in both the chord and description columns, so multi-alias bindings such as `ctrl+2` … `ctrl+0` stay visible.
- New input behavior is filed as PROPOSED in `docs/design/keybinding-decisions.md` (trigger: bare ctrl, 400ms) pending owner sign-off; `docs/KEYBINDINGS.md` + `docs/TROUBLESHOOTING.md` updated; patch changeset included.
- Regression coverage for the CSI-u wire format flag 8 turns on: composer typing, embedded-terminal forwarding, paste (`test/render/kitty-input-paths.test.tsx`, `test/tui/terminal-keys-pure.test.ts`).

## UX choice: engine-tab focus

The guide opens while an embedded engine terminal is focused. It is a transient discovery hint rather than a modal keymap: the next non-modifier key closes the guide and continues through normal Rove dispatch or PTY forwarding. Keeping it available there teaches Rove-owned escape, tab, split, and terminal-management keys at the moment users need them without intercepting engine input.

## Verification

- `tsc --noEmit` ✓
- `bun run lint` ✓
- Focused Vitest shortcut-reveal suite: 5 tests ✓
- Bun/OpenTUI render suite for the guide: 7 tests ✓, including a failing-before regression for the `ctrl+2` … `ctrl+0` clipping
- Fixed 1280×800 `/harness` → xterm.js → PTY → real OpenTUI capture with injected kitty Ctrl press/release bytes:
  - sidebar focus shows `n`, `q`, `j`/`k`, `r`, `d`, and the remaining reachable direct bindings
  - a real Claude engine tab shows workspace, session, split, and embedded-terminal direct bindings
- Full GitHub CI is green. The first `typecheck-and-test` attempt was canceled by the runner during an otherwise passing unit suite; rerunning that job completed lint, typecheck, unit tests, and build successfully.


Sma1lboy:
Heads-up: the `useKittyKeyboard: { events, allKeysAsEscapes }` request this PR added crashes iTerm2 3.5 when a Chinese input method commits text (owner report, another machine, reproducible: iTerm2 quits, not Rove). #808 reverts the flag set to `{}`; the ctrl-hold state machine, CSI-u re-encoding and modifier filtering all stay. If you want the guide back, it needs a path that does not request flag 8 on iTerm2 — e.g. request it only when `TERM_PROGRAM` is kitty/ghostty/wezterm, or after a probe shows the terminal handles a `CSI 0 u` text event. Please do not re-enable it unconditionally.

PR #754, opened 2026-09-01
feat: set a task's status from the TUI and show it on the sidebar row

Sma1lboy:
## What / why

`docs/CONCEPTS.md` promised "Each Task has a `status` you set yourself", and that was false in the TUI: nothing rendered `task.status`, and nothing outside `rove api set-status` could change it. `orch.setStatus` was fully wired end to end, with exactly one non-CLI caller — the automatic `backlog → in_progress` rule in `monitor/status-rules.ts`. Rove's own injected system prompt tells every engine to `set-status … in_review` and then says the rest is "the user's decision" — a decision the user had no UI for.

Two additions close it:

- **`statusChip(task)`** in `row-view.ts`, beside `prCheckChip`: `◇` in review, `◆` done, `†` canceled, `×` error. `backlog` and `in_progress` return null — those are the states Rove drives itself, so a mark there would only say "this row is ordinary". It renders LEFT of the PR chip: the status is what a person said, the check is what CI reports, and they can disagree.
- **"Set status"** in the sidebar's right-click menu → a picker over the six values → one `task.status` RPC (`setStatusFlow`). No new chord: placement is the owner's call (AGENTS.md), so the menu carries it alone until one is agreed, and `tree-menu.ts`'s "every entry mirrors a chord" note now records that exception instead of quietly contradicting it.

The edit is **cosmetic and stays that way**: `canceled` relabels the task and leaves its worktree, branch and sessions exactly where they were — the framing `cli/api/verbs-edit.ts` already carries, now asserted as an absence (no other orchestrator method may fire).

Glyphs deliberately reuse the tree's existing failure vocabulary (`×` errored, `†` gone) rather than inventing a second one. `◇`/`◆` pair outline/filled the way the tab rows already use `○`/`●`; `fc-list :charset=25c6` puts `◆` in Fira Code, FiraCode/JetBrainsMono Nerd Font, Menlo and Monaco — wider coverage than the `◇` already shipped for the subagent prefix. The font-vocabulary guard in `sidebar-row-state.test.ts` only scanned the state glyphs, so a chip glyph could enter unchecked; it now scans both chip blocks too.

`visual:shot` gained an `rclick:X,Y` token. The menu has no chord, so a keyboard-driven shot can never open one — without it the context-menu screenshots below are not takeable at all.

## Pass criteria — commands and real output

Isolated scratch daemon throughout (`KOBE_VISUAL_PORT_BASE=5883`, home under `.scratch/opentui-visual-5883/home`, socket/pty/web all pinned there; verified with `ps eww`). The owner's `~/.rove` and `packages/kobe/.dev-sandbox` were never touched.

**1. API round-trip + live repaint**

```
$ rove api set-status --task-id 01M1FKWTVDEETZ8RH3NB1CCE4F --status in_review
{}
$ rove api get-task --task-id 01M1FKWTVDEETZ8RH3NB1CCE4F --pretty
    "status": "in_review",
```
The sidebar repainted with `◇` with no restart — captured in `after-set-row.png`, where the toast reads `Status → in_review` and the row it was set on already wears the chip.

**2. BEFORE / AFTER screenshots**

BEFORE is `origin/main` in a separate worktree, against its own fixture seeded to `in_review`, so both sides show the same task state and only the code differs.

| | |
|---|---|
| `/Users/jacksonc/i/kobe/.scratch/status/before-row.png` | task is `in_review`; row shows **no glyph** |
| `/Users/jacksonc/i/kobe/.scratch/status/before-menu.png` | menu ends Open → … → Reorder row → Delete; **no Set status** |
| `/Users/jacksonc/i/kobe/.scratch/status/after-row.png` | `Visual Fixture` (in_review) wears `◇`; `ship-the-chip` (backlog) stays bare |
| `/Users/jacksonc/i/kobe/.scratch/status/after-menu.png` | **Set status** between Reorder row and Delete |
| `/Users/jacksonc/i/kobe/.scratch/status/after-picker.png` | all six statuses, cursor on the task's current one, marked `current` |
| `/Users/jacksonc/i/kobe/.scratch/status/after-set-row.png` | set through the menu → row repaints + `Status → in_review` toast |

`/Users/jacksonc/i/kobe/.scratch/status/api-evidence.txt` holds the raw command output.

**3. The edit stayed cosmetic** — `get-task` before vs after setting the status *through the menu*, on a task with a real materialized worktree:

```
9c9   <     "status": "backlog",     >     "status": "in_review",
14c14 <     "updatedAt": "…:12:53"   >     "updatedAt": "…:13:51"

worktreePath equal: True -> …/worktrees/fixture-repo-d1b75187148b/caiman
branch equal:       True -> ship-the-chip
tabs equal:         True -> []
changed task fields: ['status', 'updatedAt']
```

**4. Mutation** — three sites, each verified red then restored:

- delete `await ctx.orch.setStatus(taskId, next)` → 3 red in `task-actions-rename.test.ts`, naming the field: `expected "spy" to be called with arguments: [ 't1', 'in_review' ]`
- delete the `in_review` case from `statusChip` → `sidebar-row-state.golden.txt` red **and** the `worktree-status-chip` render frame red
- drop `pickStatus` from the host context → 2 red in `host-task-actions-set-status.test.ts` (the adapter seam: a flow with no picker returns cleanly and writes nothing)

**5. Gates**

- `bun x tsc --noEmit` — clean
- `bun run lint` (repo root) — `Checked 1314 files. No fixes applied.`
- `bun test test/render` — `368 pass, 0 fail`
- `bun run scripts/check-i18n.ts` — `734 keys × 2 locales in sync`
- `bun run test:fast` — `4093 passed`, 4 failing. Those 4 are the checkout LOCATION, not this change: they are `project-eligibility` path-shape rejections, and running the same two files on **unmodified `origin/main`** reproduces them — `'roveInternal'` from `~/.rove/worktrees/…`, `'temporary'` from `/tmp/…`. This diff touches neither file. CI runs from a normal path.

## Tests

- `test/golden/sidebar-row-state.golden.txt` — new `board status chip` block over the whole `TaskStatus` union, so the two deliberate silences are pinned and a seventh status shows up in the diff on the day it lands.
- `test/render/golden/worktree-status-chip.frame.txt` — new scene; a real OpenTUI frame showing `◇ ✓` cluster ordering, all four glyphs, and a bare `backlog` row. No other frame changed.
- `test/render/status-picker-dialog.test.tsx` — real keypresses; every assertion is on the value that came OUT, since an off-by-one produces a dialog that looks perfect and writes the neighbouring status.
- `test/render/sidebar-tree-menu.test.tsx` — a real right-click through the renderer proving the menu row dispatches, not just renders.
- `test/tui-react/host-task-actions-set-status.test.ts` — the adapter seam.

## Note for review

`tree-menu.ts` documents "a row's menu is what that row's KEYBOARD already does". `setStatus` is the first entry that breaks it, and deliberately: a chord needs owner sign-off, and the alternative was leaving the promise in CONCEPTS.md false for another round. The module note now says so. **If you'd rather it wait for a chord, say the word and I'll pull the menu entry and keep the chip.**

file-size-exemption: packages/kobe/src/tui-react/workspace/host.tsx — was exactly 500 lines before this change; the diff adds one JSX prop (`onSetStatusRequest`). The one real seam left in it (`startWorkspaceHost`, the process-boot half) needs edits to `src/tui/index.tsx` and `test/tui/start-tui.test.ts`, both outside this change's slice — happy to do it as a follow-up.

### DIFF
--- a/.changeset/task-status-chip-and-picker.md
+++ b/.changeset/task-status-chip-and-picker.md
@@ -1,0 +1,10 @@
+---
+"@sma1lboy/rove": patch
+---
+
+Set a task's status from the TUI, and see it on the row. The sidebar's
+right-click menu gains **Set status**, a picker over the six task statuses, and
+a task row now carries a mark once its status leaves `backlog`/`in_progress`:
+`◇` in review, `◆` done, `†` canceled, `×` error. The status is still just a
+label — picking `canceled` relabels the task and leaves its worktree, branch
+and running sessions exactly where they were.
--- a/docs/CONCEPTS.md
+++ b/docs/CONCEPTS.md
@@ -39,7 +39,14 @@ worktree), the shell folds into that Task as a new terminal tab, running
 session and all, instead of becoming a duplicate row.
 
 Each Task has a `status` you set yourself (`backlog`, `in_progress`,
-`in_review`, `done`, `canceled`, `error`).
+`in_review`, `done`, `canceled`, `error`) — from the sidebar row's right-click
+menu (**Set status**) or with `rove api set-status`. It is a label and nothing
+more: `canceled` does not stop a session or remove a worktree, and `done` does
+not close anything. Rove moves a Task from `backlog` to `in_progress` by itself
+when its engine starts a turn, and the system prompt asks the agent to set
+`in_review` when it finishes; everything past that is yours. The sidebar row
+shows the status as a mark once it leaves `backlog`/`in_progress`
+(see [TUI](./TUI.md#status-glyphs-in-the-sidebar)).
 
 **Delete is explicit and kind-aware.** A project-main Task cannot go through
 Task deletion; pressing `d` on its row instead forgets the saved project and
--- a/docs/KEYBINDINGS.md
+++ b/docs/KEYBINDINGS.md
@@ -158,8 +158,10 @@ Right-click any row for its context menu; `j`/`k` and `⏎` drive it, and a pres
 anywhere else, or `esc`, dismisses it. Common row actions also have direct
 chords. A Task or tab row also offers **New conversation** (the `ctrl+e`
 engine/shell picker) and **New shell** (a bare shell tab) for that worktree,
-both enter the Task first, exactly as pressing the chord there would. (If right-click opens your *terminal's* menu instead, see
-[Troubleshooting](./TROUBLESHOOTING.md).)
+both enter the Task first, exactly as pressing the chord there would.
+**Set status** is the one entry with no chord: it opens a picker over the six
+Task statuses and writes the one you choose. (If right-click opens your
+*terminal's* menu instead, see [Troubleshooting](./TROUBLESHOOTING.md).)
 
 ## Terminal scrollback
 
--- a/docs/TUI.md
+++ b/docs/TUI.md
@@ -44,8 +44,17 @@ Task rows carry worktree-level facts:
 |---|---|
 | `▴` | Pinned Task |
 | `+N` / `−N` | Changed and deleted files in the worktree |
+| `◇` / `◆` / `†` / `×` | Status is `in_review`, `done`, `canceled`, or `error` |
 | `✓` / `✗` / `•` | Pull-request checks passing, failing, or pending |
 | jump digit | The `ctrl+2` … `ctrl+0` shortcut currently assigned to this row |
+
+The status mark is what a person said about the Task; the check mark next to it
+is what CI reports, so the two can disagree. `backlog` and `in_progress` show
+nothing — those are the states Rove moves a Task through on its own, so a mark
+there would say only that the row is ordinary. Set the status from the row's
+right-click menu (**Set status**) or with `rove api set-status`; it is a label,
+and changing it leaves the worktree, the branch, and every running session
+alone.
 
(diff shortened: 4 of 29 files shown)


Sma1lboy:
One thing I found while wiring this, worth a decision rather than a silent choice.

`TaskEditor.setStatus` refuses `done ↔ error` with `IllegalTransitionError`, and its comment says the guard exists "to surface bad code". That was true while `rove api set-status` and the auto-status rule were the only callers. The picker offers all six values, so a person can now walk into it: `done` → pick `Error` → a toast reading `Couldn't set status: illegal transition for task 01M1F…: done -> error`, with a raw ULID in it.

Nothing breaks — the flow catches and reports, and `task-actions-rename.test.ts` now pins that. But a tripwire for programmer error has quietly become a user-facing rule, and there are three ways to go:

1. **Leave it** (what's here). The toast is honest, the case is rare.
2. **Grey the option out in the picker** when the current status makes it illegal — but that puts the orchestrator's rule in a second place to keep in sync.
3. **Drop the guard.** The statuses are cosmetic labels; refusing one relabel while allowing `done → canceled → error` is a rule that only half exists.

I didn't pick — the guard isn't mine and 2 and 3 both reach outside this change. Happy to do whichever in a follow-up.

PR #812, opened 2026-09-03
A recycled tab keeps its pinned engine, gets a fresh id, and ctrl+a w closes it

Sma1lboy:
Batch C items **C1** and **C2** — they share the tab core, so one PR. Full proof (screenshots, mutation runs, pixel diffs, gate numbers) is in the first PR comment.

**C1 — `recycleTabs` dropped the tab's pinned engine and reused `tab-1`.**
It returned `initialTabs()` wholesale with only `title`/`autoTitle` copied, so a tab you pointed at another engine respawned on the *task's* engine while still wearing the old conversation's title — and the new tab was `tab-1` again, breaking the `TabBase.id` "never reused within a task" invariant that Inbox episodes and orphan suppression both key on. It now carries `vendor`/`engineCommand` the way its sibling `reopenHintFor`/`reopenTabs` already does and mints the id from `state.nextOrdinal`.

**C1 — `ctrl+a` `w` was a dead key on an unsplit tab.**
`workspace.split.close` was the sole owner of `prefixKeys: ["w"]` and only registers while split; `chat.tab.close` — the row that is live when it is not — declared none, so the second stroke hit no enabled prefix binding and was consumed. `chat.tab.close` now carries the same `prefixKeys: ["w"]` its mutually-gated twin already had. **No chord added or moved** — it is the existing chord's existing documented meaning, and nothing becomes reachable that `ctrl+w` did not already reach. Flagging it for owner sign-off anyway; recorded in `docs/design/keybinding-decisions.md`, and `docs/KEYBINDINGS.md:55` now matches.

**C2 — one request-box factory.**
Seven copies of the pending-slot/take/take-unclaimed reducer in `terminal-tabs-shared.ts` collapse into one `requestBox<T>()`. Exported names, argument shapes, the shared listener set, the consumption order in `use-tab-requests.ts` and the deliberate absence of `takeUnclaimed` for four of the seven are all unchanged, so no call site moves. 441 → 387 lines.

**CI follow-up (this push).** `terminal-tabs-core.ts` had crossed the size cap, so the seam it was hiding is now named: everything that mints a *whole* `TabsState` from something other than a user action — first mount, restart snapshot, reviving an emptied task, recycling on exit — moves to `terminal-tabs-lifecycle.ts`. Those four have to agree about ordinal minting and carried identity, and this bug is what a silent disagreement between two of them looks like. `TabsState` joins the other shapes in `terminal-tab-shapes.ts` so the new module needs no runtime import back into core; core re-exports, so no importer changes. The test file splits the same way. 502 → 365 and 523 → 427.

### Gates
`bun run typecheck`, `bun run lint`, `bash scripts/file-size-check.sh` — clean. `bun test test/render` — 431 pass, 0 fail. `bun run test:fast` — 4261 pass, 6 fail, all pre-existing and environmental (identical failure set on my other, disjoint branch): `continue-live-vendor` reads the owner's locally-registered `claudecpa` engine out of the real `state.json`, and `project-eligibility` / `open-dir-cmd` are the known `~/.rove/worktrees` path-shape false failures. CI runs from an ordinary Linux checkout.

coverage-exemption: packages/kobe/src/tui-react/workspace/TerminalTabs.tsx — measured baseline on origin/main is LF 263 / LH 90 = 34.22%, already 16pp under the floor; this PR's two touched lines (one deleted import, one `prefixAction` wrapper) leave it at LF 265 / LH 90 = 33.96%. Nothing in test/render mounts TerminalTabs at all — `show-workspace-close-last.test.tsx` documents why ("opens a daemon socket and needs four more providers"), so the whole `useBindings` block containing the touched line is uncovered on both sides. The changed behaviour is proven instead through the repo's own ground truth for it: a `/harness` BEFORE/AFTER pair driving the real mounted component with real keys (see the comment).

### DIFF
--- a/.changeset/tab-recycle-keeps-its-engine.md
+++ b/.changeset/tab-recycle-keeps-its-engine.md
@@ -1,0 +1,5 @@
+---
+"@sma1lboy/rove": patch
+---
+
+A tab that exits as the task's last one now comes back as the engine it was pinned to. Recycling used to hand back a bare first tab, so a tab you pointed at Codex respawned as the task's engine while still wearing the Codex conversation's title. It also reused `tab-1`, which let a recycled tab inherit a dead tab's Inbox episodes and its in-flight orphan suppression — ids are minted from the task's ordinal counter now, and are never reused. Separately, `ctrl+a` `w` closes the tab when there is no split, instead of being swallowed.
--- a/docs/KEYBINDINGS.md
+++ b/docs/KEYBINDINGS.md
@@ -52,7 +52,7 @@ shows only actions that can run right now.
 | `ctrl+a` `o` | Open the Task directory in your editor |
 | `ctrl+a` `m` | Reorder sidebar rows (scope-aware: tab / task / project) |
 | `ctrl+a` `n` | New task |
-| `ctrl+a` `w` | Close the active split |
+| `ctrl+a` `w` | Close the active split, otherwise the tab — the prefix twin of `ctrl+w` |
 | `ctrl+a` `1` / `2` / `3` | Kanban / Routines / Issues |
 | `ctrl+a` `z` | Toggle zen mode |
 | `ctrl+a` `,` | Open Settings |
--- a/docs/design/keybinding-decisions.md
+++ b/docs/design/keybinding-decisions.md
@@ -422,6 +422,27 @@ answer `setStatus` and the two copies got: the capability exists now, and a
 chord can be added later if reaching for the mouse turns out to be the
 friction rather than the safety.
 
+## `<prefix> w` closes the tab when there is no split
+
+No chord was added or moved. `ctrl+w` and `<prefix> w` are one documented
+gesture — "close the active split, otherwise the tab" — split across two
+registry rows that are mutually gated so exactly one is live:
+`workspace.split.close` while the active tab is split, `chat.tab.close`
+otherwise.
+
+Only the split row declared `prefixKeys: ["w"]`. On an unsplit tab the second
+stroke therefore found no enabled prefix binding, and the dispatcher CONSUMES
+a prefix miss on purpose (so a deliberate sequence can never type into a
+terminal). The key was eaten, silently, while `KEYBINDINGS.md`, the row's own
+F1 help text and its i18n twin all promised it closed the tab.
+
+The fix gives `chat.tab.close` the same `prefixKeys: ["w"]` its twin already
+had, so whichever row the gate leaves live owns BOTH strokes. Nothing new is
+reachable that the direct chord did not already reach, and no other binding
+loses a key. Its handler is wrapped in `prefixAction` for the same reason the
+split row's is: the prefix HUD's clickable option resolves through `action`,
+not `cmd`.
+
 ## Adding or moving a chord
 
 Get owner sign-off on direct versus prefix placement, the selected key, and
--- a/packages/kobe/src/tui-react/workspace/TerminalTabs.tsx
+++ b/packages/kobe/src/tui-react/workspace/TerminalTabs.tsx
@@ -59,7 +59,6 @@ import {
   initialShellTabs,
   initialTabs,
   isTabSplit,
-  recycleTabs,
   rehydrateTabs,
   renameActiveTab,
   selectTab,
@@ -425,7 +424,9 @@ export function TerminalTabs(props: TerminalTabsProps): ReactNode {
   useBindings(() => ({
     enabled: props.focused && !activeIsSplit,
     bindings: bindByIds({
-      "chat.tab.close": () => tabClose.closeActive(),
+      // prefixAction so the prefix HUD's clickable option resolves too —
+      // the mouse path needs `action`, not just `cmd`.
+      "chat.tab.close": prefixAction(() => tabClose.closeActive()),
       "chat.tab.rename": requestRename,
     }),
   }))
--- a/packages/kobe/src/tui-react/workspace/terminal-tabs-shared.ts
+++ b/packages/kobe/src/tui-react/workspace/terminal-tabs-shared.ts
@@ -130,42 +130,71 @@ export function knownTaskTabs(
 }
 
 /**
- * Cross-component "activate this tab" request (the F7 attention jump). The
- * mounted TerminalTabs for `taskId` consumes it via the listener; a task
- * that isn't mounted yet consumes it on mount (the host selects the task
- * first, TerminalTabs mounts, then reads the pending request). Unknown tab
- * ids are dropped on consume — the tab may have closed meanwhile.
+ * One cross-component request slot: a caller (the sidebar, a plugin, the F7
+ * jump) names a task, and whoever owns that task's tab state claims it.
+ *
+ * Every box shares ONE listener set — a request of any kind wakes every
+ * mounted consumer, which is what lets `use-tab-requests.ts` drain them all in
+ * a fixed order on a single pass. `take` hands the payload only to the task it
+ * was addressed to. `takeUnclaimed` returns it whoever it was for, and is
+ * re-exported ONLY by the three boxes whose callers have a background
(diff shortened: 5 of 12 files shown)


Sma1lboy:
Batch C, items **C1** and **C2** (they share `terminal-tabs-shared.ts` / the tab core, so one PR).

Screenshots live at `/Users/jacksonc/i/kobe/.scratch/loop-r1/shots-batchC/` — all captured through the only ground-truth path (`/harness` browser → xterm.js → PTY sidecar → real OpenTUI), BEFORE on `origin/main`, AFTER on this branch.

The fixture has no logged-in engines, so the recycle repro uses two **sentinel engines** installed into the isolated fixture home: `engineCommand.claude` and a custom `sentinel` engine each point at a one-line script that prints which engine it is and exits into the worktree shell. The recycled tab then says out loud which engine it actually spawned.

---

## C1 — tab recycle drops the pinned engine, reuses `tab-1`, and `ctrl+a w` is a dead key

### 5a + 5b — `recycleTabs`

`recycleTabs(prev)` returned `initialTabs()` wholesale with only `title`/`autoTitle` copied over. Two consequences:

* the tab's pinned `vendor` / `engineCommand` were dropped, so a tab the user pointed at another engine respawned via `engineTabSpawn`'s fallback on the **task's** engine — invisibly, because the carried title still advertised the old conversation;
* the new tab was `tab-1` again and `nextOrdinal` rewound to `2`, breaking the `TabBase.id` "never reused within a task" invariant that Inbox episodes (`taskTabExists`) and orphan suppression (`noteClosedPtyKey`, keyed `${taskId}::tab-1`) both depend on.

It now carries the pinned engine the way its sibling `reopenHintFor`/`reopenTabs` already does, and mints the id from `state.nextOrdinal` (signature is `recycleTabs(state, prev)`; the single call site in `use-tab-close.ts` passes `stateRef.current`). The `if (fresh.activeId === active.id) bumpResetToken()` guard went with it — the id always changes now, so the `${taskId}::${tabId}` pty key changes and `Terminal` re-acquires on its own.

**PASS criterion:** a unit test on `recycleTabs` proving vendor **and** a never-reused id, plus a harness BEFORE/AFTER pair.

**Proof — unit tests** (`test/tui/terminal-tabs-core.test.ts`), verified to fail against the old implementation:

```
# with recycleTabs reverted to the old body:
 × recycleTabs carries the exited tab's pinned vendor and engineCommand
 × recycleTabs mints a fresh id from nextOrdinal instead of reusing tab-1
 Tests  2 failed | 30 passed (32)

# on this branch:
 ✓ test/tui/terminal-tabs-core.test.ts (32 tests)
```

**Proof — harness.** Journey: enter the task (tab-1, task engine) → `ctrl+e` → pick `sentinel` (tab-2, pinned) → close tab-1 from the tree row menu → `exit` in the surviving tab, which is the recycle branch of `handleActiveExit`.

* `c1-recycle-before.png` — the recycled tab reads **`ENGINE LAUNCHED: claude (task default)`** and the tree shows it as `sentinel 1`: wrong engine, old title, ordinal rewound to 1 while a `2` row still exists.
* `c1-recycle-after.png` — **`ENGINE LAUNCHED: sentinel (tab-pinned)`**, tree row `shell 3`.

The fixture's persisted tab state after the AFTER run confirms both halves directly:

```json
{"tabs":[{"kind":"engine","id":"tab-3","ordinal":3,"title":null,"vendor":"sentinel","sessionId":null},
         {"kind":"engine","id":"tab-2","title":null,"ordinal":2,"spawned":true}],
 "activeId":"tab-3","nextOrdinal":4}
```

### 5c — `ctrl+a` `w` on an unsplit tab

`workspace.split.close` was the sole owner of `prefixKeys: ["w"]` and is only registered while the tab is split. `chat.tab.close` — the row that IS live when it is not — declared no prefix keys, so the second stroke found no enabled prefix binding and `keymap-dispatch.ts` consumed it on purpose (a prefix miss must never type into a PTY). `docs/KEYBINDINGS.md:55`, the split row's own F1 help text ("Close active split (tab when unsplit)") and its i18n twin all promised otherwise.

**No chord was added or moved.** `ctrl+w` / `<prefix> w` is one documented gesture split across two mutually-gated registry rows; the fix gives `chat.tab.close` the same `prefixKeys: ["w"]` its twin already had, so whichever row the gate leaves live owns *both* strokes. Nothing becomes reachable that the direct chord did not already reach, and no other binding loses a key. Its handler is now wrapped in `prefixAction` for the same reason the split row's is — the prefix HUD's clickable option resolves through `action`, not `cmd`. Recorded in `docs/design/keybinding-decisions.md`, and `docs/KEYBINDINGS.md:55` now matches the behaviour. **Flagging it explicitly for owner sign-off** even though it is the existing chord's existing meaning.

**PASS criterion:** in the harness, `ctrl+a` then `w` on an unsplit tab closes the tab.

* `c1-prefixw-before.png` — prefix HUD reads `ctrl+a + w ∅` (no action), tab still there.
* `c1-prefixw-after.png` — HUD reads `ctrl+a + w → Close chat t…`, the tab is gone and the pane shows "No sessions here".

---

## C2 — one request-box factory

Seven copies of the "one pending slot, sweep the shared listener set, consume-if-my-task" reducer in `terminal-tabs-shared.ts` collapse into one `requestBox<T>()`. Preserved deliberately:

* all seven exported `request*` / `take*` names and argument shapes, so **no call site changes**;
* the single shared `tabActivationListeners` set — any request still wakes every mounted consumer;
* the fixed consumption order in `use-tab-requests.ts` (activation, open, pane-close, new-tab, adopt, move, close);
* the **absence** of `takeUnclaimed` for activation / open / new-tab / pane-close. Only close, adopt and move have background fallbacks; the factory's docblock now states why once instead of four times.

Where the exported signature already is `(taskId, payload)`, the wrapper is the box method itself (`export const requestTabActivation = activationBox.request`); the three positional-argument callers (`requestTabOpen`, `requestPaneClose`, `requestTabMove`) and the three `takeUnclaimed*` readers keep real wrappers.

**Not split into `terminal-tabs-requests.ts`:** the brief said to only if the file stayed near 500 lines. It went 441 → 387.

**PASS criterion:** fast + render green, sidebar "New conversation" flow identical, net line reduction.

* `bun run test:fast` — see the note below; `test/plugins/tab-open.test.ts` green.
* `bun test test/render` — **431 pass, 0 fail** across 99 files, including `test/render/tab-requests-new-tab.test.tsx`.
* `c2-newconv-before.png` / `c2-newconv-after.png` — AFTER captured on a fixture rebuilt with `KOBE_VISUAL_FRESH=1` as the brief requires. A pixel diff of the dialog region `(350,220)-(910,595)` reports **0 differing rows of 375**. The two shots differ only outside it: the fixture's randomized worktree directory name in the shell prompt (`mamba` vs `walrus`) and the tree row's live process-title probe (`shell 1` vs `claude 1`) — both fixture artefacts of the rebuild, not of this change.
* Net: `terminal-tabs-shared.ts` 441 → 387 lines (−54); non-comment, non-blank lines 239 → 203 (−36).

---

## Gates

* `bun run typecheck` — clean, repo root.
* `bun run lint` — clean, repo root (1357 files).
* `bun test test/render` — 431 pass, 0 fail.
* `bun run test:fast` — 4261 pass, **6 fail**, all pre-existing and environmental, none in the code this PR touches:
  * `test/tui/continue-live-vendor.test.ts` (2) — asserts `"claude"` but receives `"claudecpa"`, the owner's locally-registered custom engine leaking in from the real `state.json`.
  * `test/state/project-eligibility.test.ts` (2) and `test/cli/open-dir-cmd.test.ts` (2) — the known false failures from running inside `~/.rove/worktrees`, where the path shape is classified `roveInternal`/`temporary`.

  I could not produce a clean local baseline for these without creating a worktree I am not allowed to remove, so I am flagging them rather than claiming them green; CI runs from an ordinary Linux checkout and should be the arbiter.

PR #814, opened 2026-09-03
Batch B: worktree removal, PTY host lifecycle, PR-status staleness, dead task.reorder slice

Sma1lboy:
Batch B from loop-r1: daemon and orchestrator lifecycle. Four items, one PR — B3/B4 were meant to be a separate PR, but their commits also fix a line in `test/render/pty-freeze.test.ts` that B2 introduces, so splitting would mean rewriting history for no review benefit.

---

## B1 — a task whose worktree is nested in its repo was permanently undeletable

`packages/kobe/src/orchestrator/worktree/manager-remove.ts`

`git worktree remove` does two jobs — deregister the metadata, delete the directory — and they fail apart. The classification of that split asked `rev-parse --git-common-dir` from *inside* the worktree, and git's discovery walks up parent directories. For a worktree nested under its own repo it resolved the **parent** and reported a path git had already forgotten as still registered, so the removal threw, every retry threw identically, and the task parked in `deletion.phase: "error"` forever. Rove creates exactly that layout for every remote project (`<checkout>/.rove/worktrees/<slug>`) and every legacy repo-local root.

Registration is now asked of the **owning repo** via `git worktree list --porcelain` (`isRegisteredWorktree`), at both the ownership gate and the post-failure classification. The vanished-directory prune also probed the repo with `cwd` set to the missing path — which cannot spawn — so it never ran; it probes the parent now.

**PASS: `rove api delete --wait --force` converges on a half-failed removal, and the task leaves `rove api list`.**

Reproduced first with real git (`rev-parse` returns `../../../.git`, exit 0, from a deregistered nested worktree). Then end to end against an isolated daemon, same scenario on both binaries — worktree deregistered, `.git` pointer unlinked (the Linux shape), an undeletable `chmod 555` directory left behind:

```
--- BEFORE: rove api delete --wait --force ---
{"status":"failed","error":"... exited with code 128: fatal: '<wt>' is not a working tree"}
--- BEFORE: rove api delete --wait --force (RETRY) ---
{"status":"failed","error":"... exited with code 128: fatal: '<wt>' is not a working tree"}
--- BEFORE: rove api list ---
tasks still listed: ['nested residue', 'residue-BEFORE']      # deletion.phase = error, forever

--- AFTER: rove api delete --wait --force ---
{"status":"removed"}
--- AFTER: rove api delete --wait --force (RETRY) ---
{"status":"not_found"}
--- AFTER: rove api list ---
residue-AFTER is gone
```

The undeletable file survived both runs (`keep.txt=precious`) — residue is reported, never deleted. The daemon's own audit line, unprompted:

```
daemon [task-deletion-audit]: removed task ... — git deregistered the worktree but could NOT
delete <wt> (a previous removal deregistered the worktree). Nothing further is needed in Rove
```

Full transcripts: `/tmp/loop-r1-b/b1-before.txt`, `/tmp/loop-r1-b/b1-after.txt`.

Tests: `test/orchestrator/worktree-remove-partial.test.ts` gains a nested-layout block (3 cases). Red before the fix: the retry case and the vanished-directory prune both fail; the first-call case is red only where git unlinks the pointer (Linux CI) and its comment says so. All 8 existing cases stay green (11/11).

**No harness screenshots** — this is a daemon/CLI path with no rendered surface. The CLI output above and the audit line are the ground truth.

---

## B2 — PTY host: closed sessions came back, deaths re-fired, parked state was clobbered

`packages/kobe-daemon/src/daemon/{pty-host.ts,pty-exit-watch.ts}`

- **4a** `kill()` dropped the freeze record synchronously and then awaited `endChild`; the `onExit` handler force-froze the same session a tick later, recreating the file with a fresh `updatedAt` that outlived the TTL prune. `maybeFreeze` now honours `closedByRequest` — one guard in the shared writer, covering both call sites.
- **4b** the exit watcher keyed `seen` by `record.key` while the store keys engine records `<key>#engine`, so its prune dropped each engine record on the sweep that created it and every later sweep re-fired the death. Keyed by store key now.
- **4c** `detachClient` ran `applyParkedOnDetach` over every session, and its `sinks.size !== 0` guard passes for an already-parked one — so the daemon's own 15s liveness poll wiped every tab's parked bookkeeping. Gated on `sinks.delete(token)`.

**PASS: a closed session must not come back in the next host incarnation.**

The two incarnations `pty-server.ts` actually runs (`kill` → `loadFrozenSessions` → `restoreFrozen`), which is exactly what `rove api pty-list` reads:

```
### BEFORE (unfixed pty-host)
incarnation 1, session open   : [ "task-a::tab-1" ]
incarnation 1, after close    : []
incarnation 2, `pty-list` shows: [ "task-a::tab-1" ]     # the closed tab, resurrected

### AFTER (fixed pty-host)
incarnation 2, `pty-list` shows: []
```

Repeated inside the **real visual-harness fixture**, closing a tab through the same `kill()` path and restarting its pty host:

```
BEFORE:  records left after close: [ "<task>::tab-9" ]
         [pty-host pty] restored 1 frozen session(s) from disk
AFTER:   records left after close: []
         restore lines: 0
```

Transcript: `/tmp/loop-r1-b/b2-resurrection.txt`.

Tests, all red before / green after:
- `test/render/pty-freeze.test.ts` — open a real `/bin/cat` PTY, `await host.kill(...)`, assert the real file store holds nothing. (Asserting before the await is green either way; the await is the test.)
- `test/daemon/pty-exit-watch.test.ts` — one engine-layer record, two sweeps, fires exactly once. Was 2.
- `test/render/pty-host.test.ts` — an unrelated socket close leaves a parked session's `parkedSessions` / `parkedScreenBytes` intact.

**No harness screenshot.** I tried: planted the record the unfixed `kill` leaves into the fixture's freeze store, restarted the harness, and confirmed the host logged `restored 1 frozen session(s) from disk` — but the fixture's single task has never been entered and so has no tab rows for the orphan-tab backstop to hang a ghost row on. Rather than hand you a picture that shows nothing, the evidence is the two transcripts above plus the three red/green tests.

---

## B3 — a PR chip that stopped updating now says so

`pr-status-collector.ts`, `task-editor.ts`, `tui/panes/sidebar/row-view.ts`

`TaskPRStatus.lastError` was declared, persisted, mirrored into daemon contracts and deliberately excluded from `samePrStatus` — and never assigned. A failing `gh` poll logged to the daemon log, backed off, and left the chip showing its last reading with nothing on screen to say it was stale.

Three parts: the collector writes `lastError` on both failure branches and clears it on any poll that reaches the provider (including `empty`); the sidebar chip keeps its glyph and drops its colour while the marker stands; and `setPRStatus`'s diff guard checks `lastError` alongside `samePrStatus`. **That third part is the one the brief did not name and the fix does not work without it** — `samePrStatus` omits `lastError` on purpose so a healthy PR does not churn a write per tick, and `setPRStatus` reused it wholesale, so every attempt to set or clear the marker read as a redundant call and was silently dropped. I only found it because the e2e below still showed no `lastError` after the collector change.

**PASS: `PATH=/nonexistent` daemon restart → `rove api get-task` shows `prStatus.lastError`.**

Against the isolated daemon, with an SSE subscriber held open so the poller actually ticks (it skips passes with no GUI attached):

```
daemon [pr-status-poller]: gh pr list failed (missing-binary) for task ... — keeping last PR status, backing off

$ rove api get-task ... | .task.prStatus
{ "checkState": "passing", "number": 42, ..., "lastError": "missing-binary" }
```

**Harness screenshots** — `/harness` → xterm.js → PTY sidecar → real OpenTUI, same seeded task whose poll is failing, same viewport, only `row-view.ts` differing:

| | |
|---|---|
| BEFORE | `/tmp/loop-r1-b/shots/b3-before.png` · crop `b3-before-crop.png` — `feat/pr-chip ✓` in healthy **green** while polling is broken |
| AFTER | `/tmp/loop-r1-b/shots/b3-after.png` · crop `b3-after-crop.png` — same row, same `✓`, **muted**: the fact has not changed, but nothing is confirming it |

Tests: 5 new cases in `test/daemon/pr-status-poller.test.ts` (set on failure while keeping the last value; no re-persist across a repeated failure; clear on an unchanged successful poll; clear on `empty`; no marker for a task that never had a PR status). Three are red before the fix. The golden `sidebar-row-state.golden.txt` gains a `stale=0/1` cross — the three chip states each get a muted twin, the three no-chip states are unchanged. Read the diff before merging.

---

## B4 — deleted the dead `task.reorder` vertical slice

Zero producers, zero consumers. It ordered cards in the web board's per-status columns; the board no longer carries task cards (`kobe-web/src/lib/board.ts:164`) and the sidebar's move mode uses `task.move`. Removed: the protocol verb, the handler, `reorderTasks` in contracts/core/task-editor, `reorder` in the store, the `position` field, both serializers, the codec branch, the web type, and the test references — including `test/orchestrator/reorder.test.ts`, which tested `store.reorder` and nothing else. **That file deletion is the only one in this PR;** it goes with the method it covers, under the loop's standing instruction to remove this slice.

**PASS: zero hits, and an existing `tasks.json` carrying `position` still loads.**

```
$ grep -rn "task.reorder\|reorderTasks\|store.reorder" packages/ --include='*.ts' --include='*.tsx'
(exit 1 — zero hits)
```

Not assumed — measured. Wrote `position` onto all three tasks of a real `tasks.json`, restarted the isolated daemon:

```
wrote position onto 3 tasks
--- rove api list against a tasks.json carrying `position` ---
loaded 3 tasks: ['nested residue', 'residue-BEFORE', 'residue-BEFORE2']
position surfaced? False
```

Pure deletion, so no screenshots.

---

## Gates

| | |
|---|---|
| repo-root `bun run lint` | clean (1356 files) |
| `bun run typecheck` | clean |
| `bun test test/render` | see below |
| `KOBE_INCLUDE_SOCKET=1 bun run test:socket` | see below |
| `bun run test:fast` | 4255 passed, **6 failed** |

Those 6 are environmental, not mine. Identical set on `origin/main` checked out in this same worktree:

```
FAIL test/cli/open-dir-cmd.test.ts (2)          — path-shape rejection under ~/.rove/worktrees
FAIL test/state/project-eligibility.test.ts (2) — expected 'roveInternal' to be null, same cause
FAIL test/tui/continue-live-vendor.test.ts (2)  — reads this machine's registered `claudecpa` engine
```

## Not proven

- No harness screenshots for B1 (no rendered surface) or B2 (fixture has no tab rows for the ghost to appear on) — both are argued above with their real evidence instead.
- B2's 4c PASS criterion asked for `rove doctor` reporting the right parked count under a live liveness poll. I proved the mechanism with a unit test against the real `PtyHost` (`stats()` / `list()` after an unrelated `detachClient`) rather than standing up a background poll against a running host.


Sma1lboy:
## Owed harness evidence for this PR (b1 + b2), recaptured cleanly

The partial PNGs from the interrupted run (`/tmp/loop-r1-b/shots/`) were
unreliable — `b2-after-crop.png` was byte-identical to `b1-before-crop.png` —
so nothing from them is reused. Everything below was captured fresh from two
independent trees and two independent isolated fixtures:

- **BEFORE** = `.scratch/e4-before`, detached at `62264f292` (`72f00420a^`,
  the commit before this PR), harness on `KOBE_VISUAL_PORT_BASE=6373`
- **AFTER** = `.scratch/e4-after`, detached at `origin/main` `c94011d42`,
  harness on `KOBE_VISUAL_PORT_BASE=6473`

Both through the ground-truth path: `/harness` → xterm.js → PTY sidecar → real
OpenTUI, against a real daemon and a real fixture repo.

---

### b1 — nested worktree undeletable

Scenario, identical on both sides (`/tmp/loop-r2-e/e4-b1.sh`): create a
worktree **nested inside its own repo** at the legacy repo-local path
`<repo>/.rove/worktrees/<slug>`, adopt it as a task, then half-fail the
removal — `git worktree remove --force` to deregister it, unlink the `.git`
pointer, recreate the directory with a file in it, and `chmod 555` the
directory so the delete cannot finish. Then `rove api delete --wait --force`
twice.

| | |
| --- | --- |
| BEFORE — the `nested-before` row is still in the sidebar after two forced deletes | `/tmp/loop-r2-e/e4/b1-before.png` · crop `/tmp/loop-r2-e/e4/b1-before-crop.png` |
| AFTER — the row is gone | `/tmp/loop-r2-e/e4/b1-after.png` · crop `/tmp/loop-r2-e/e4/b1-after-crop.png` |

The `rove api delete` JSON behind those two frames (full transcripts:
`/tmp/loop-r2-e/e4/b1-before.txt`, `/tmp/loop-r2-e/e4/b1-after.txt`):

```
BEFORE  delete #1  {"queued":true,"status":"failed","error":"failed to remove worktree for task
                    01M1KEDGVWYYGAS4XCJPSKG3TF: git worktree remove --force …/fixture-repo/.rove/
                    worktrees/nested-before … exited with code 128: fatal: '…/nested-before' is
                    not a working tree"}
BEFORE  delete #2  {"queued":true,"status":"failed","error":  … identical …}
BEFORE  api list   ['Visual Fixture', 'nested residue']      ← the task never leaves

AFTER   delete #1  {"queued":true,"status":"removed"}
AFTER   delete #2  {"queued":false,"status":"not_found"}
AFTER   api list   ['Visual Fixture']
```

The undeletable `chmod 555` directory survives on both sides — the fix is that
the task stops being hostage to it, not that the file gets forced away.

---

### b2 — closed session comes back after host restart

Scenario, identical on both sides: enter the fixture task in the harness so it
has a tab row, close that tab from the tab row's context menu (**Close tab**),
then stop and restart the visual servers **without** `KOBE_VISUAL_FRESH`, so
the fixture and its on-disk PTY state persist.

| | |
| --- | --- |
| BEFORE — after the restart, PTY host boots and restores the closed session | `/tmp/loop-r2-e/e4/b2-before.png` · crop `/tmp/loop-r2-e/e4/b2-before-crop.png` |
| AFTER — nothing restored | `/tmp/loop-r2-e/e4/b2-after.png` · crop `/tmp/loop-r2-e/e4/b2-after-crop.png` |

The log line, from each side's own `<fixture home>/.rove/pty.log`:

```
BEFORE  [pty-host pty] restored 1 frozen session(s) from disk
AFTER   (no such line — 0 matches across 3 host boots)
```

and the freeze store on disk immediately after the tab was closed:

```
BEFORE  ls <home>/.rove/pty-sessions/
        01M1KE9R1AZ290HPP20N4HF3CJ%3A%3Atab-1.json     ← a closed tab left a record
AFTER   ls <home>/.rove/pty-sessions/
        (empty)
```

**What I could NOT reproduce, and exactly where it blocks.** The brief expected
the BEFORE frame to show a *ghost tab row* in the sidebar. It does not, and the
step that blocks is the row's source: closing the tab through the row menu
removes the task's persisted tab snapshot, so after the restart the sidebar has
no tab row to draw — the resurrection is real but it lives one layer down, in
the PTY host's in-memory session map, and the TUI's orphan-tab backstop does not
surface it for a task whose snapshot was cleanly removed. (The previous worker
hit the same wall from the other direction: they planted a freeze record by hand
and found the fixture task had never been entered, so it had no tab rows at
all.) The two BEFORE/AFTER sidebar frames are therefore *identical by design* —
the honest visible evidence is the log line and the freeze-store listing above,
which differ exactly where the fix does. Both frames are attached anyway so the
absence is on the record rather than asserted.

Supporting captures kept alongside: `/tmp/loop-r2-e/e4/b2-probe-crop.png` (the
`claude 1` tab row present after entering the task),
`/tmp/loop-r2-e/e4/b2-closing-crop.png` (the row menu with **Close tab**),
`/tmp/loop-r2-e/e4/b2-closed-crop.png` and
`/tmp/loop-r2-e/e4/b2-after-closing-crop.png` (the tab row gone on each side).

PR #815, opened 2026-09-03
One owner for the sidebar callbacks, and for engine display names

Sma1lboy:
Batch C items **C3** and **C4** — two "one owner" refactors, no behaviour change. Detailed proof in the first comment.

**C3 — `SidebarTaskCallbacks` was re-declared at four more layers.** The type's own docblock says it exists so "the two surfaces can't drift", and then `HostSidebarProps` spelled all fourteen callbacks out inline, `host-sidebar.tsx` re-listed the same fourteen into a `common` object, and `use-tree-bindings.ts` declared four of them a fourth time with the parameter renamed `id` instead of `taskId`.

`HostSidebarProps` now extends the shared type. Required-ness is preserved exactly, not weakened: `Readonly<Required<Omit<SidebarTaskCallbacks, "onLandRequest">>>` plus `Readonly<Pick<…, "onLandRequest">>`, because the host supplied every callback except that one. The `common` object is gone — `HostSidebarProps` is `SidebarTreeProps` plus the rail's own two props, so the tree takes the rest by spread instead of a re-listing of thirty names that can silently fall out of date. `TreeBindingsOpts` takes a `Pick` of the four it uses, which also settles the `id`/`taskId` split. No callback changed from menu-only to chord-bound or back.

**C4 — engine display names had three owners.** `humanizeSlug` was duplicated verbatim (settings model + daemon settings adapter, differing only in equivalent spellings), and the adapter's `engineLabel` was a third derivation of the same answer that `engineDisplayName` already gives from the registry. Both copies are deleted; `humanizeSlug` now lives in `engine/interactive-command.ts` beside `engineDisplayName`, which sits below both `src/core/` and `src/tui/` so neither import inverts. It stays a **pure string transform** — it is the fallback written *into* `engineName.<id>`, so giving it the override lookup would overwrite a user-typed name with the title-cased slug on the next write. `engineLabel` folds into `engineDisplayName(id)`. The bare `"claude"` fallbacks in the adapter (default engine, and the two resets when the active custom engine is deleted) and in `handlers-add.ts` now read a named `DEFAULT_VENDOR` from `types/vendor.ts`. `kobe-web/src/lib/vendor.ts` deliberately untouched — that package is frozen for new work.

### Gates
- `bun run typecheck` clean, **no new `any` or non-null assertion** at any of the five C3 sites; `bun run lint` clean; `bash scripts/file-size-check.sh` clean.
- `bun test test/render` — 431 pass, 0 fail (`sidebar-tree-menu.test.tsx`, `sidebar-tree.test.tsx`, `sidebar-row-chords-cursor.test.tsx` included).
- Render coverage gate clean on both touched React files: `host-sidebar.tsx` 90.20%, `use-engine-settings.ts` 81.41%.
- Named unit tests green: `sidebar-tree-menu-items`, `sidebar-controller`, `daemon-settings-adapter`, `settings-rows`, `interactive-command-overrides`, and the `engine-owned-copy` architecture guard.
- **C4 byte-identity:** the daemon settings snapshot for a state holding a custom engine id with **no** `engineName` override — the exact path both deleted copies served — is byte-for-byte identical before and after (953 bytes, `cmp` clean), including `"label":"my-local-agent"`.
- **C3 harness:** sidebar row menu open, BEFORE on `origin/main` and AFTER on this branch through `/harness` → xterm.js → PTY sidecar → real OpenTUI. Pixel diff of the menu region `(28,158)-(204,400)`: **0 differing rows of 242**.
- `bun run test:fast` — 4259 pass, 6 fail, all pre-existing and environmental; the identical failure set appears on my other, disjoint branch (#812). `continue-live-vendor` reads the owner's locally-registered `claudecpa` engine out of the real `state.json`; `project-eligibility` / `open-dir-cmd` are the known `~/.rove/worktrees` path-shape false failures. CI runs from an ordinary Linux checkout.

Net: −108 lines across ten files.

### DIFF
--- a/.changeset/one-owner-for-engine-names.md
+++ b/.changeset/one-owner-for-engine-names.md
@@ -1,0 +1,5 @@
+---
+"@sma1lboy/rove": patch
+---
+
+Engine display names now come from one place. `humanizeSlug` was implemented twice (settings model and the daemon settings adapter) and `engineLabel` was a third derivation of the same answer; the registry-backed `engineDisplayName` in `engine/interactive-command.ts` owns it now, and the hard-coded `"claude"` fallbacks in the adapter and `add` handler read a named `DEFAULT_VENDOR`. No user-visible change — the daemon settings snapshot is byte-identical.
--- a/packages/kobe/src/cli/api/handlers-add.ts
+++ b/packages/kobe/src/cli/api/handlers-add.ts
@@ -20,7 +20,7 @@ import type { SerializedTask } from "@sma1lboy/kobe-daemon/daemon/protocol"
 import { resolveCommandProtocol } from "../../engine/engine-presets.ts"
 import { ulid } from "../../orchestrator/index/ulid.ts"
 import type { TaskStatus } from "../../types/task.ts"
-import type { VendorId } from "../../types/vendor.ts"
+import { DEFAULT_VENDOR, type VendorId } from "../../types/vendor.ts"
 import type { DaemonRpc } from "../daemon-session.ts"
 import { dispatcherEnvPayload, withPeerProvenance } from "./dispatcher.ts"
 import { FANOUT_CAP, buildCountPlan, parseAgentsSpec } from "./flags.ts"
@@ -222,7 +222,7 @@ async function addParallel(
   const choice = await engineChoice(ctx, repo)
   const plan: VendorId[] = agentsSpec
     ? parseAgentsSpec(agentsSpec)
-    : buildCountPlan(count ?? 1, choice.vendor ?? "claude")
+    : buildCountPlan(count ?? 1, choice.vendor ?? DEFAULT_VENDOR)
   if (plan.length > FANOUT_CAP) {
     throw new ApiError(
       `a parallel round of ${plan.length} exceeds the cap of ${FANOUT_CAP} — spawn in batches`,
--- a/packages/kobe/src/core/daemon-settings-adapter.ts
+++ b/packages/kobe/src/core/daemon-settings-adapter.ts
@@ -3,6 +3,7 @@ import {
   engineCommandKey,
   engineDisplayName,
   engineNameKey,
+  humanizeSlug,
 } from "../engine/interactive-command.ts"
 import { AUTO_STATUS_KEY } from "../state/auto-status.ts"
 import { DISPATCHER_KEY } from "../state/dispatcher.ts"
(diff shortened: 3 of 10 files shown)


Sma1lboy:
## Proof detail

### C3 — screenshots

Files (absolute paths on the dispatcher's box):
* `/Users/jacksonc/i/kobe/.scratch/loop-r1/shots-batchC/c3-rowmenu-before.png` — captured on `origin/main`
* `/Users/jacksonc/i/kobe/.scratch/loop-r1/shots-batchC/c3-rowmenu-after.png` — captured on this branch

Journey both times: click the `visual-fixture` row → right-click it → screenshot the row menu. Same isolated harness fixture recipe both times.

Pixel comparison of the menu, decoded straight from the PNGs:

```
sidebar row-menu region (28,158)-(204,400): differing rows = 0 of 242
```

The two shots differ only outside that region — the fixture's randomized worktree directory name in the shell prompt (`mamba` vs `mouse`), which changes on every `KOBE_VISUAL_FRESH=1` rebuild.

### C4 — byte-identical daemon settings snapshot

Fixture `state.json` (a custom engine with **no** name override, and a second one with a name, so both branches of the deleted `engineLabel` are exercised):

```json
{
  "activeTheme": "claude",
  "defaultVendor": "my-local-agent",
  "customEngineIds": ["my-local-agent", "named-agent"],
  "engineCommand.my-local-agent": "my-local-agent --interactive",
  "engineName.named-agent": "Bespoke Agent",
  "engineCommand.named-agent": "bespoke run"
}
```

`daemonSettingsSnapshot()` was invoked against that home twice — once with `origin/main`'s `daemon-settings-adapter.ts` restored in place, once with this branch's — and the two response bodies compared with `cmp`:

```
BYTE-IDENTICAL ✓ (953 bytes)
```

The `engines[]` entry that matters:

```json
{"id":"my-local-agent","label":"my-local-agent","command":"my-local-agent --interactive",
 "isBuiltin":false,"isCustom":true,"isDefault":true}
```

`engineLabel(state, id)` was `state["engineName."+id] || engineDisplayName(id)`, and `engineDisplayName` reads that same key from that same `state.json` via `getPersistedString`, so the fold is an identity by construction — this run confirms it end to end rather than by inspection.

### C4 — where `"claude"` was replaced, and where it was not

Replaced (vendor fallbacks): `daemon-settings-adapter.ts` `lastSelectedVendor` default, and the `defaultVendor` / `lastSelectedVendor` resets when the active custom engine is deleted; `handlers-add.ts` `buildCountPlan`. **Not** replaced: `activeTheme: stringValue(state.activeTheme, "claude")` in the same function — that literal is a *theme* name that happens to share the spelling, and `DEFAULT_VENDOR` there would be a type-checking bug that reads as a fix.

### C3 — required-ness

`HostSidebarProps` had thirteen of the fourteen callbacks required and `onLandRequest?` optional, while `SidebarTaskCallbacks` has all fifteen members (fourteen callbacks + `moveMode`) optional. `Readonly<Required<Omit<…, "onLandRequest">>> & Readonly<Pick<…, "onLandRequest">>` reproduces that exactly, including `moveMode: boolean` staying required and the `readonly` modifiers the interface declared. `host.tsx` supplies all fourteen and still typechecks unchanged.

PR #837, opened 2026-09-03
fix: require the web token on the PTY sidecar's routes

Sma1lboy:
## M1 — the PTY WebSocket was the one route the web token did not gate, and it is the one that executes commands

**PASS criterion:** a transcript against a `rove web` on an isolated home — BEFORE a tokenless upgrade connects and `/pty/send` accepts a body; AFTER both are refused (401) and the same requests with the token succeed. Plus a browser screenshot of `/task/<id>` showing a live terminal before and after.

### What changed

- **`packages/kobe-web/pty-auth.mjs`** (new) — the gate. Reads the same `0600` token file the daemon mints (`<home>/.rove/web-token`, legacy `.kobe` fallback) rather than taking the secret from whoever spawned the sidecar; it starts from four places (`rove web`, `dev.ts`, Playwright, by hand) and a launcher that forgot to forward it would silently reopen the hole. Constant-time compare, fail closed.
- **`pty-server.mjs`** — `ws /pty`, `POST /pty/send`, `POST /pty/close` all require the token (`ws` refuses a rejected upgrade with 401). Origin check kept as defence in depth. `fetchSpec` now presents the token on its own `/api/*` hop to the daemon.
- **`src/lib/terminal.ts`** — `ptyUrl` goes through `withWebTokenQuery`, the same channel the `/events` SSE stream uses (a WebSocket cannot set a request header).
- **`dev.ts` / `fixture-core.ts` / e2e fixtures** — Vite, not the daemon, serves the SPA in the dev and capture stacks, so the `<meta>` tag never appears there; `VITE_ROVE_WEB_TOKEN` is the browser's channel. `dev.ts` sets it from the token file; the visual/hero fixtures pin `FIXTURE_WEB_TOKEN` written before their daemon starts, and their teardown `/pty/close` calls now carry it.

### Proof — live isolated `rove web` (not read-verified)

Isolated home `/tmp/m1-pty/home`, `rove web --port 45874`, sidecar on 45876, every inherited control (`KOBE_DAEMON_SOCKET_PATH`, `ROVE_INVOKED_AS`, …) unset.

**BEFORE** (`b1c5519ef`, main):

```
ws  attach   (no token)     : CONNECTED, output="[?1034hprobe$ "
POST /pty/send  (no token)  : HTTP 200 {"sent":true,"spawned":false}
POST /pty/close (no token)  : HTTP 200 {"closed":true}
ws  attach   (wrong token)  : CONNECTED, output="[?1034hprobe$ "
```

A tokenless upgrade returns a live shell prompt, a tokenless `/pty/send` types into it, a tokenless `/pty/close` kills it, and a *wrong* token works just as well.

**AFTER** (this branch):

```
ws  attach   (no token)     : REFUSED http 401
POST /pty/send  (no token)  : HTTP 401
POST /pty/close (no token)  : HTTP 401
ws  attach   (token=…)      : CONNECTED, output="[?1034hprobe$ "
POST /pty/send  (bearer)    : HTTP 200 {"sent":true,"spawned":true}
POST /pty/close (bearer)    : HTTP 200 {"closed":true}
ws  attach   (wrong token)  : REFUSED http 401
```

Full transcripts: `/Users/jacksonc/i/kobe/.scratch/m1-pty/before.txt` and `after.txt`.

### Proof — the browser (Playwright, `/task/<id>` on the real dashboard)

Same isolated stack, entry navigation carrying `?token=` exactly as `rove web` prints it, then the in-app route change, then a shell tab opened in the task worktree.

- BEFORE — `/Users/jacksonc/i/kobe/.scratch/m1-pty/before.png`: the terminal reads **`failed to start shell: unauthorized: this request carried no valid web token`** followed by `[detached: spawn failed — reattach below]`.
- AFTER — `/Users/jacksonc/i/kobe/.scratch/m1-pty/after.png`: a live prompt, `jacksonc@Jacksons-MacBook-Pro avocet %`, in the task's own worktree.

### A regression this uncovered

The web terminal has been **dead since 0.9.60**, not merely ungated. The sidecar fetches each tab's launch spec from the daemon's `/api/terminal-spec` / `/api/engine-spec`; that route began requiring the token in 0.9.60 (#692) and the sidecar was never taught to send one. Verified directly against the isolated daemon:

```
$ curl -s -o /dev/null -w "%{http_code}\n" "http://127.0.0.1:45874/api/terminal-spec?taskId=nope"
401
$ curl -s -H "authorization: Bearer $TOK" "http://127.0.0.1:45874/api/terminal-spec?taskId=nope"
{"error":"task not found: nope"}
```

That is why the BEFORE screenshot shows a broken terminal rather than a working one. The visual harness never caught it because `KOBE_PTY_DEV_COMMAND` short-circuits `fetchSpec` before it reaches the daemon. The fix is the same credential and the same file as the gate, so it ships here; the changeset says so.

### Tests

`test/pty-auth.test.ts` (new, 12 cases) plus one case in `test/pty-url.test.ts`. Mutation-verified — reverting `withWebTokenQuery` in `ptyUrl` and flipping `ptyTokenAccepted`'s fail-closed guard turns 4 of them red:

```
FAIL test/pty-auth.test.ts > ptyTokenAccepted > refuses an absent token on either side …
FAIL test/pty-auth.test.ts > expectedPtyToken > stays empty when no token file exists …
FAIL test/pty-auth.test.ts > ptyRequestAuthorized > refuses the tokenless and wrong-token callers …
FAIL test/pty-url.test.ts  > ptyUrl > carries the web token
Tests  4 failed | 11 passed (15)
```

### Gates

| gate | result |
|---|---|
| repo-root `bun run lint` | pass (1397 files) |
| `bun run test:fast` | 4336 passed, 6 pre-existing failures (below) |
| `bun test test/render` | 455 pass, 0 fail |
| `KOBE_INCLUDE_SOCKET=1 bun run test:socket` | 789 passed (94 files) |
| `packages/kobe-web` `bun run test` | 589 passed (73 files) |

The 6 `test:fast` failures are not from this branch. Measured on a clean `origin/main` worktree under `~/i`:

- `test/cli/open-dir-cmd.test.ts` (2) and `test/state/project-eligibility.test.ts` (2) **pass** there — they fail here only because this worktree lives under `~/.rove/worktrees`, whose path shape those tests classify as rove-internal.
- `test/tui/continue-live-vendor.test.ts` (2) **fails on pristine `origin/main` too** — pre-existing red on main, unrelated to this change.

### Not proven

The hero/atlas capture stacks received the same token plumbing by inspection; I ran neither, since they need real engine credentials. The visual fixture path uses the same code.

### DIFF
--- a/.changeset/pty-websocket-requires-the-web-token.md
+++ b/.changeset/pty-websocket-requires-the-web-token.md
@@ -1,0 +1,34 @@
+---
+"@sma1lboy/rove": patch
+---
+
+Require the web bearer token on the PTY sidecar, and let the sidecar present
+it to the daemon.
+
+The PTY routes were the one part of the web surface the token never reached,
+and they are the part that runs commands: `ws /pty` spawns the task's engine or
+a shell in its worktree, `POST /pty/send` types into that process, and `POST
+/pty/close` kills it. Their only gate was the Origin check, which admits a
+request with **no** `Origin` — every non-browser client — and admits **any**
+loopback origin, so a page the user happened to have open on another localhost
+port could attach as well. WebSocket upgrades get no CORS preflight, so nothing
+else stood in the way. Against a running `rove web`, a tokenless upgrade
+returned a live shell prompt and a tokenless `/pty/send` typed into it; a wrong
+token worked just as well. Exposure was limited to whatever could reach the
+sidecar's loopback port — on a shared machine, any other local user, which is
+the boundary the token file's 0600 mode exists to hold — and not reachable from
+a remote network unless the port had been bound off loopback.
+
+All three routes now require the token that every REST and SSE caller already
+sends: the browser puts it on the WebSocket URL the way the `/events` stream
+does, and the sidecar refuses an upgrade without it (`401`). The Origin check
+stays as defence in depth. The sidecar reads the same `0600` token file the
+daemon mints rather than taking the secret from whoever spawned it, so a
+launcher that forgot to forward it cannot silently reopen the gap; no readable
+file means no request is served.
+
+This also repairs the web terminal, dead since 0.9.60: the sidecar fetches each
+tab's launch spec from the daemon's `/api/*`, that route began requiring the
+token in 0.9.60, and the sidecar was never taught to send one — so every engine
+and shell tab died with `failed to start shell: unauthorized: this request
+carried no valid web token` before any PTY was spawned.
--- a/docs/design/web-dashboard.md
+++ b/docs/design/web-dashboard.md
@@ -226,10 +226,10 @@ unconditional.
 Both the daemon web transport and the PTY sidecar
 ([`pty-server.mjs`](../../packages/kobe-web/pty-server.mjs)) bind
 **`127.0.0.1` by default**. `KOBE_WEB_HOST` overrides only when a LAN bind is
-intended. A PTY WS is arbitrary command exec in the worktree, so the upgrade enforces a
-**localhost-Origin allowlist** (`localhost`/`127.0.0.1`/`[::1]`) — a browser
-cross-origin upgrade is rejected; a non-browser client (no `Origin`) is allowed
-since there's no ambient browser session to ride.
+intended. A PTY WS is arbitrary command exec in the worktree, so the upgrade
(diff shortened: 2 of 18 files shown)


Sma1lboy:

PR #843, opened 2026-09-03
test: make the PTY sidecar's daemon spec fetch reachable from a test

Sma1lboy:
## O2 — the PTY sidecar's daemon fetch has no test, because the harness never runs it

`packages/kobe-web/pty-server.mjs` exports nothing and starts an HTTP server at
import time, so `fetchSpec` was unreachable from any test. Meanwhile every
runner that touches the sidecar sets `KOBE_PTY_DEV_COMMAND`
(`playwright.config.ts:67`, `e2e/visual-serve.ts:32`, `e2e/hero-serve.ts:38`)
and the function returns before the daemon hop. `pty-session-lifecycle.test.ts`
injects its own `fetchSpec`, so it covers the manager, not the route choice,
not the bearer header #837 added, not the error shaping. That is how a web
terminal broken from 0.9.60 to 0.9.102 passed ~40 CI runs.

### What changed

| | |
| --- | --- |
| `packages/kobe-web/pty-spec.mjs` (new) | `createSpecFetcher({ port, readToken, env, fetchImpl })` — collaborators as parameters, the way `pty-auth` / `pty-env` / `pty-scrollback` / `pty-session-lifecycle` / `origin-policy` already take them. The dev-command branch stays in the same module. |
| `pty-server.mjs` | `const fetchSpec = createSpecFetcher({ port: DAEMON_WEB_PORT })`; 26 lines out. |
| `packages/kobe-web/test/pty-spec.test.ts` (new) | 8 cases driving the real branch with an injected `fetch`. |
| `use-host-sessions.ts` | `sameSessions` exported + tested (its poll is off under every runner, so the comparator never executed in any track). |
| `doctor-report.ts` | drops `KOBE_TEST_ENGINE` — no code in either package reads it. |
| `docs/HARNESS.md` | new "What the harness replaces, and therefore cannot prove" section. |
| `.github/workflows/ci.yml` | runs `kobe-web`'s vitest suite — see the note at the bottom. |

`readToken` is a function rather than a captured value: the daemon may mint the
token file after the sidecar starts, and `expectedPtyToken` memoizes only once
non-empty, so a token read at construction time would wedge the gate shut for
the process lifetime.

### PASS criterion and proof

**Criterion:** the new test is RED when the bearer header is dropped and GREEN
as written; `packages/kobe-web` `bun run test` green; `docs/HARNESS.md` gains
the blind-spot section; `bun run visual` unaffected; `KOBE_TEST_ENGINE` gone.

Mutation runs — three separate sites, each restored afterwards:

```
=== MUTATION 1: `headers: token ? {authorization: `Bearer ${token}`} : {}` → `headers: {}` ===
     × carries the bearer token — without it a real tab dies unauthorized before any PTY spawns
     × re-reads the token per call, so a file minted after startup is picked up
      Tests  2 failed | 6 passed (8)

=== MUTATION 2: route choice collapsed to "/api/engine-spec" ===
     × asks the engine route in engine mode and the terminal route in shell mode
      Tests  1 failed | 7 passed (8)

=== MUTATION 3: `!res.ok ||` dropped from the error gate ===
     × falls back to the status form when a failing response says nothing
      Tests  1 failed | 7 passed (8)

=== RESTORED ===
      Tests  8 passed (8)
```

`sameSessions`, same treatment:

```
=== MUTATION: title comparison dropped ===   × reports a change on each field the tree reads
=== MUTATION: length guard dropped ===       × reports a change when a session appears or disappears
=== RESTORED ===                             4 pass
```

Suites (this branch, macOS):

```
bun run lint                     Checked 1400 files. No fixes applied.
bun run typecheck                Exited with code 0
bun run --filter kobe-web test   Test Files 74 passed (74) | Tests 597 passed (597)
bun test test/render             459 pass, 0 fail (104 files)
bun run visual                   5 passed (14.9s)   ← harness path undisturbed
```

Coverage gates, both green with no exemption:

```
BASE_REF=main bun scripts/coverage-gate.mjs
  ok packages/kobe/src/cli/doctor-report.ts — 100%
BASE_REF=main KOBE_RENDER_COVERAGE=1 bun scripts/coverage-gate.mjs
  ok packages/kobe/src/tui-react/panes/sidebar/use-host-sessions.ts — 57.58%   (was 40.63%, floor 50%)
```

`grep -rn "KOBE_TEST_ENGINE" --exclude-dir=node_modules .` → one hit, the
comment in `packages/kobe/src/env.ts:5`.

**`bun run test:fast` on this branch reports 6 failures** — `continue-live-vendor`
×2 and `project-eligibility` / `open-dir-cmd` ×4. Those are the pre-existing
environment-shaped failures this checkout's path and the developer's
`state.json` cause (they are 0 on CI's clean Linux runner, and the sibling PR
for batch O1 removes them everywhere). None are in files this PR touches.

### Two deviations from the brief, both deliberate

1. **`sameSessions`'s test is on the render track, not the fast track.** The
   brief said fast track. `use-host-sessions.ts` value-imports `react`, so
   `coverage-gate.mjs`'s `isRenderTrackOnly` classifies it render-track-only —
   a vitest test scores nothing against it and would leave the file at 40.63%,
   under the 50% floor, forcing an exemption for a file that can simply be
   covered. `test/render/*.test.ts` already hosts pure-logic tests
   (`mention-text.test.ts`).
2. **CI now runs `kobe-web`'s vitest suite** (4 lines in `typecheck-and-test`).
   Not in the brief; I added it because without it this PR's whole point is
   undone: root `lint`/`typecheck`/`test` all skip `kobe-web` and
   `visual-ground-truth` runs Playwright, so those 74 vitest files — including
   the new one — gate nothing. It costs ~2s. Note `kobe-web`'s own `bun run
   check` is red on `main` for 5 pre-existing format/import issues in
   `src/lib/{markdown,settings,store}.ts`, so only the test step was added, not
   the lint step. Drop the step if you'd rather take it separately.

No screenshots: no rendered surface changes — the diff is a module extraction,
one deleted env-dump row, one added `export`, tests, docs and CI.


Sma1lboy:
## Behavior job: one red run, then green on rerun — not this PR

First run of `behavior` failed 3 tests, all three retries each:

```
× pure-tui-unread-restart      expected undefined to be 'tab-1'          (:179)
× pure-tui-live-tab-title      B's snapshot never listed tab-1           (:126)
× pure-tui-adopt-orphan-tab    expected undefined to be defined          (:77)
```

Rerun of the same job on the same commit: **pass (1m33s)**. A rerun going green
only proves instability, so here is what rules this PR out:

- **The base is green.** `main` @ `337cd58b4` — this branch's exact parent —
  passed `behavior` on CI 20 minutes earlier.
- **The sibling PR is green.** #844 (batch O1, same window, same runner image)
  passed `behavior` first try.
- **Locally green on this branch alone**, not just on the combined tree:
  `git checkout test/pty-sidecar-spec-fetch && bun run build && bun run
  test:behavior` → `Test Files 11 passed (11) | Tests 32 passed (32)`.
- **The diff cannot reach that path.** The behavior track spawns the built kobe
  CLI; this PR's only `src` edits are one `export` keyword on a pure function
  and one deleted string in the doctor env dump. `pty-spec.mjs` is the
  kobe-web sidecar, which the behavior track never loads.

**Where it probably does come from:** all three failures are "the TUI's
`terminalTabs` snapshot never reached state.json", and all three of these files
were rewritten four commits ago by #836 (`Exit on a readiness signal, and make
the behavior harness wait for it`) — which replaced `child.kill()` + a flat 5s
wait with `closeTui` resolving on the child's `exit` event. That is exactly the
state-flush ordering that failed here, and it has less than a day of CI history.
Worth a look independently of this PR; `--retry=2` means all three attempts land
on the same runner, so a retry does not rescue it.

PR #845, opened 2026-09-03
Floor every relative-time step from one module; collapse git-snapshot's four spawns into one

Sma1lboy:
Batch P from `.scratch/loop-r7/briefs.md` — briefs 3 and 4. Two commits, reviewable
independently; the first is a behavior change with screenshots, the second is a
behavior-preserving refactor with an identity check.

Screenshots and JSON captures are archived at
`/Users/jacksonc/.rove/worktrees/kobe-0aff3858ab76/snipe/.scratch/loop-r7-p/shots/`
(also `/tmp/loop-r7-p/`).

---

## P1 — two relative-time rules, one clock: floor is the owner

**What changed.** `relativeBuckets` (`packages/kobe/src/lib/relative-time.ts`) rounded
at every step while `relativeAgeMs` (`tui/history/message-core.ts`) floored, so the
same instant read two ways depending on which screen you were on.
`relativeBuckets` now floors and owns the `3m`/`2h`/`4d` label too, as `relativeAge`;
`relativeAgeMs` is deleted and its six consumers (worktrees page, attention inbox ×2,
issue events, plugins section, and `relativeTime`) import from the one module. The
header comment that recorded floor-vs-round as undecided now states the decision and
the two reasons for it.

**PASS criterion.** Harness BEFORE/AFTER on the Routines page and the Inbox, on a
fixture whose next run is 100 minutes out and whose newest run event is ~100 seconds
old: BEFORE `in 2h` / `2m ago`, AFTER `in 1h` / `1m ago`, with the Inbox reading the
same event as `1m` throughout. Plus `grep -rn relativeAgeMs packages/` → 0, the
existing relative-time tests updated deliberately, and fast + render green.

**Proof.**

| | BEFORE | AFTER |
| --- | --- | --- |
| Routines | `shots/routines-before.png` | `shots/routines-after.png` |
| Inbox | `shots/inbox-before.png` | `shots/inbox-after.png` |

Read off the images — three labels move, all on the Routines page:

| surface | BEFORE | AFTER |
| --- | --- | --- |
| next-run preview, 100 min out | `in 2h` | `in 1h` |
| run list, event ~100 s old | `#2 dispatched 2m ago` | `#2 dispatched 1m ago` |
| run list, event 100 min old | `#1 dispatched 2h ago` | `#1 dispatched 1h ago` |
| Inbox, the same ~100 s event | `1m` | `1m` (unchanged — it already floored) |

The Inbox pair is the point of the change: BEFORE, the Routines run list called a
100-second-old event `2m ago` while the Inbox called it `1m`. AFTER, they agree.

Captured through `/harness` → xterm.js → PTY sidecar → real OpenTUI, on a FRESH
fixture each side (`visual:serve` restarted between the two after `bun run build`), with
`KOBE_VISUAL_PORT_BASE=5873`. The fixture is seeded by
`.scratch/loop-r7-p/seed-routine.ts`, which writes the fixture daemon's
`automations.json` + `attention-inbox.json` with offsets measured from the seed instant
and restarts that daemon so its store re-reads them.

```
$ grep -rn "Math.round" packages/kobe/src/lib/relative-time.ts    # exit 1, no matches
$ grep -rn relativeAgeMs packages/kobe/src packages/kobe/test packages/kobe-web/src
(no matches)
```

The only surviving `relativeAgeMs` string in the repo is `packages/kobe/CHANGELOG.md:2709`,
a shipped release note describing history — left alone.

`relativeAge` keeps exactly one arithmetic step of its own (seconds); minutes/hours/days
come from `relativeBuckets`.

**Test changes, deliberate.**
- `test/lib/relative-time.test.ts` — the test named "each step ROUNDS — the behavior
  both consumers shipped with" (an explicit placeholder) becomes the floor contract, and
  `relativeBuckets(23h)` moves from `days: 1` to `days: 0`. Added the two cases the
  decision turned on (100 min → 1 h, 36 h → 1 d), the 119/120-minute boundary pair, the
  59.999 s/60 s minute boundary, and the `relativeAge` cases moved out of
  `test/tui-react/plugins-core.test.ts` (they had been living next to the plugins view
  model because that was the function's last consumer).
- `test/render/automations-page-keys.test.tsx` and `test/render/failure-toast.test.tsx`
  both pinned `nextRunAt` at exactly `Date.now() + 3_600_000`. Flooring makes that read
  `59m` the moment the render takes a millisecond, so both move to 100 minutes — which
  reads `in 1h` under the decided rule and would have read `in 2h` under the old one, so
  the fixtures now distinguish the two rules instead of straddling them.

**Gates.**
```
bun run lint            ok (1399 files)
bun run typecheck       ok (4 packages)
bun test test/render    455 pass, 0 fail
bun run test:fast       4354 passed, 6 failed
```
Those 6 are environment-shaped and pre-existing — `test/state/project-eligibility` ×2 and
`test/cli/open-dir-cmd` ×2 (this checkout is under `~/.rove/worktrees`, which
`pathRejection` reads as `roveInternal`) and `test/tui/continue-live-vendor` ×2 (the real
`~/.config/rove/state.json` has `claudecpa` registered). They are brief 1's evidence and
batch O's target. Proven not mine: with only my touched files reverted to HEAD, the same
3 files produce the identical 6 failures —
```
$ git checkout HEAD -- <my 9 files>
$ bun x vitest run test/cli/open-dir-cmd.test.ts test/state/project-eligibility.test.ts \
    test/tui/continue-live-vendor.test.ts
  Test Files  3 failed (3)
       Tests  6 failed | 31 passed (37)
```

---

## P2 — `git-snapshot.ts` spawns git one way now

**What changed.** Four `spawnSync("git", …)` blocks, each with its own copy of the
options and its own `try`/`catch`, collapse into one module-private `git()` helper. The
fourth call had already drifted — no `stdio`, so stderr took the default instead of
`ignore`; nothing reads `stderr` here, so unifying it is not observable. The helper's doc
names the shape the render-path sync whitelist entry is granted for (one shot, O(refs),
2 s cap, stderr discarded), so the whitelist now points at one call site instead of four.

`spawned` is the one field that keeps this exactly behavior-preserving: the four
`catch` fallbacks were NOT identical — three treated a thrown spawn as failure, but
`hasNoCommits` returned `false` ("has commits"). Mapping a throw to `status: 1` alone
would have flipped that one. No test can reach it (`spawnSync` with constant options does
not throw), which is why it is documented rather than covered.

**PASS criterion.** `bun run test:fast` green on the covering tests; an end-to-end
identity check byte-identical before/after; net line reduction stated; and the brief's
mutation check — forcing `missing` to `false` must turn the "git is not installed" vs
"not a repo" case red.

**Proof.**

Behavior fingerprint — every exported path (`validateRepoPath`, `getCurrentBranch`,
`listLocalBranches`, `DEFAULT_BASE_REF`) across 8 cases including a dirty repo, a subdir,
an unborn HEAD, a plain file, a missing path, the empty string, and git absent from PATH
(`.scratch/loop-r7-p/git-snapshot-probe.ts`):
```
$ diff git-snapshot-before.json git-snapshot-after.json && echo IDENTICAL
IDENTICAL
$ md5 -q git-snapshot-before.json git-snapshot-after.json
c5ca35336553d0871bd18b81a44a1aa4
c5ca35336553d0871bd18b81a44a1aa4
```

`rove api get-task` and `rove api collect` for a dirty fixture worktree (the sidebar's
`+N −N` chip), rebuilt CLI on both sides — `diff` clean:
```json
{
 "taskId": "01M1M3QPCX0HNNWC6WSGCHZ653",
 "branch": "visual-fixture",
 "changes": { "added": 2, "deleted": 0 },
 "base": { "baseRef": "main", "ahead": 0, "behind": 0,
           "diff": { "files": 0, "insertions": 0, "deletions": 0 } }
}
```
(That chip is daemon-computed and never routed through `git-snapshot.ts`, so this is a
nothing-else-moved check rather than a proof of the refactor; the fingerprint above is
the proof.)

```
$ wc -l packages/kobe/src/tui/lib/git-snapshot.ts     196 -> 183
$ grep -c 'spawnSync(' packages/kobe/src/tui/lib/git-snapshot.ts     1
$ bun x vitest run test/tui/lib/git-snapshot.test.ts test/tui/render-path-sync-guard.test.ts
  Test Files  2 passed (2)   (whitelist entry untouched)
```

Net −13 lines, not the brief's estimated −23: the estimate assumed a bare helper, and
the six-line doc block recording the whitelist's granted shape is the part of this change
worth keeping.

**Mutation checks, two different sites.**
1. `missing: isBinaryMissing(out)` → `missing: false`:
   `× validateRepoPath > says git is missing, not that the folder isn't a repo, when git is absent`
   → 1 failed | 8 passed. The ENOENT distinction is genuinely covered.
2. `return !name || name === "HEAD" ? null : name` → `return name`: **stayed green.**
   The brief's rule says the test is owed before the refactor lands, so this commit adds
   it — a detached HEAD is the one case where git exits zero and still has no branch name
   (`--abbrev-ref` prints the literal `HEAD`), and prefilling `baseRef` with that string
   would aim `git worktree add` at a ref that moves. With the new test, the same mutation
   goes red:
   `× getCurrentBranch > degrades to null on a detached HEAD` → 1 failed | 9 passed.

No changeset: behavior-preserving, nothing user-visible. No screenshots: no TUI surface
changes.

**Deliberately skipped** (per the brief's own re-measurement): the daemon
`task not found` shaping. Ten one-line sites plus a four-line helper is a positive net
line change with no consistency boundary to defend.

---

## Not proved

- The 6 pre-existing failures above are left for batch O; I did not fix them.
- `spawned` in `git()` has no test, by construction (see above).
- `bun run test:socket` / `test:behavior` were not run — this batch touches neither the
  daemon protocol nor the behavior harness.

### DIFF
--- a/.changeset/relative-time-floors.md
+++ b/.changeset/relative-time-floors.md
@@ -1,0 +1,5 @@
+---
+"@sma1lboy/rove": patch
+---
+
+One relative-time clock, and it floors. The Routines page rounded while every other surface floored, so the same instant read two ways: a routine firing in 1h40m previewed as `in 2h` and fired twenty minutes early, and a run 100 seconds old showed `2m ago` on the Routines page while the Inbox called it `1m`. `relativeBuckets` in `lib/relative-time.ts` now floors and owns the label too — `relativeAgeMs` is gone from `tui/history/message-core.ts` and its six consumers read `relativeAge` from the one module. A sub-minute run also now reads `now` / `just now` instead of `1m ago`.
--- a/packages/kobe/src/lib/relative-time.ts
+++ b/packages/kobe/src/lib/relative-time.ts
@@ -1,17 +1,31 @@
 /**
- * Stepwise coarse buckets for a relative-time delta: minutes from the raw
- * millisecond delta, hours from minutes, days from hours. Both TUI consumers
- * (the Routines schedule preview's `formatRelative` and the Automations
- * page's `formatWhen`) derive their units from here so the two can't drift
- * apart — independent copies are how one screen ends up flooring while its
- * neighbor rounds.
+ * The product's one relative-time clock: coarse buckets for a millisecond
+ * delta, and the compact `3m` / `2h` / `4d` label built on them. Every TUI
+ * surface that prints an age or a countdown derives its units here — the
+ * Routines schedule preview (`formatRelative`), the Automations page
+ * (`formatWhen`), the attention inbox, the worktrees page, the issue event
+ * list, and the plugins section — so two screens cannot disagree about the
+ * same instant.
  *
- * Every step uses Math.round. Floor-vs-round is a display-convention call
- * (a countdown that floors never overstates headroom); when
- * that's decided, change it HERE so every consumer moves together.
+ * Every step FLOORS. Two reasons the rounding variant lost: a countdown that
+ * rounds up overstates headroom (a routine 1h40m out rendered `in 2h` and
+ * then fired twenty minutes early), and `relativeAge`'s seconds step has no
+ * rounding rule anyone wants (a 45-second-old event is not `1m ago`). Flooring
+ * makes both read early rather than late.
  */
 export function relativeBuckets(absMs: number): { minutes: number; hours: number; days: number } {
-  const minutes = Math.round(absMs / 60_000)
-  const hours = Math.round(minutes / 60)
-  return { minutes, hours, days: Math.round(hours / 24) }
+  const minutes = Math.floor(absMs / 60_000)
+  const hours = Math.floor(minutes / 60)
+  return { minutes, hours, days: Math.floor(hours / 24) }
+}
+
+/** Relative age of an epoch-ms timestamp ("3m", "2h", "4d"); negative deltas clamp to "0s". */
+export function relativeAge(ms: number, nowMs: number = Date.now()): string {
+  const delta = Math.max(0, nowMs - ms)
+  const secs = Math.floor(delta / 1000)
+  if (secs < 60) return `${secs}s`
+  const { minutes, hours, days } = relativeBuckets(delta)
+  if (minutes < 60) return `${minutes}m`
+  if (hours < 24) return `${hours}h`
+  return `${days}d`
 }
--- a/packages/kobe/src/tui-react/component/issue-detail-parts.tsx
+++ b/packages/kobe/src/tui-react/component/issue-detail-parts.tsx
@@ -16,7 +16,7 @@ import { useT } from "../i18n"
 import { DialogLabel } from "../ui/dialog-parts"
 import { EVENT_FEED_LIMIT, type EventRow, eventRows } from "./issue-events-core"
 
-/** Width of the age column — "999d" is the widest {@link relativeAgeMs} yield. */
+/** Width of the age column — "999d" is the widest {@link relativeAge} yield. */
 const AGE_CELLS = 4
 
 export function IssueEventsSection(props: { taskId: string; orchestrator: RemoteOrchestrator | null }) {
--- a/packages/kobe/src/tui-react/component/issue-events-core.ts
+++ b/packages/kobe/src/tui-react/component/issue-events-core.ts
@@ -9,7 +9,7 @@
  */
 
 import type { RecentTaskEvent } from "../../client/remote-orchestrator"
-import { relativeAgeMs } from "../../tui/history/message-core"
+import { relativeAge } from "../../lib/relative-time"
 import { truncateEnd } from "../../tui/lib/truncate"
 
 /** Rows the drawer shows — the tail of the ring, not the whole 100. */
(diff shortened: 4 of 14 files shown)


Sma1lboy:
P1 harness captures (PNGs live on the machine that ran them — \`gh\` cannot upload attachments, so these are paths, not embeds):

```
/Users/jacksonc/.rove/worktrees/kobe-0aff3858ab76/snipe/.scratch/loop-r7-p/shots/routines-before.png
/Users/jacksonc/.rove/worktrees/kobe-0aff3858ab76/snipe/.scratch/loop-r7-p/shots/routines-after.png
/Users/jacksonc/.rove/worktrees/kobe-0aff3858ab76/snipe/.scratch/loop-r7-p/shots/inbox-before.png
/Users/jacksonc/.rove/worktrees/kobe-0aff3858ab76/snipe/.scratch/loop-r7-p/shots/inbox-after.png
```

What they show, verbatim from the frames:

| | BEFORE | AFTER |
| --- | --- | --- |
| Routines row, next run 100 min out | `fixture-repo · 48 11 * * * in 2h` | `fixture-repo · 51 11 * * * in 1h` |
| Routines RECENT RUNS, ~100 s old | `· #2 dispatched  2m ago` | `· #2 dispatched  1m ago` |
| Routines RECENT RUNS, 100 min old | `· #1 dispatched  2h ago` | `· #1 dispatched  1h ago` |
| Inbox ATTENTION, the same ~100 s event | `✓ done  1m` | `✓ done  1m` |

The last row is why the change exists: BEFORE, one instant read `2m ago` on Routines and `1m` in the Inbox.

PR #874, opened 2026-09-04
fix: the delete gate stops reporting "nothing here" without looking

Sma1lboy:
Task AO — Batch A (A1, A2, A3) and B. C1/C2/D deliberately left for another worker.

Every item was reproduced against a real repo built and then damaged, under an isolated
daemon whose `ROVE_/KOBE_ {HOME_DIR, DAEMON_SOCKET_PATH, DAEMON_PID_PATH, PTY_SOCKET_PATH,
PTY_PID_PATH, DAEMON_WEB_PORT}` were each assigned inline (never `env $VAR`). Fixtures live
under `.scratch/loop-r15-ao/` only; nothing was run against `/Users/jacksonc/i/kobe` or
anything under `/Users/jacksonc/.rove/worktrees/`.

---

## A1 — a worktree whose only work is gitignored deleted with no gate, no confirm, no salvage

**PASS criterion:** the no-force delete is refused with `DIRTY_WORKTREE` naming `HANDOFF.md`
and `.scratch/`, and the `--force` retry writes a `refs/rove/salvage/…` whose `ls-tree` lists
both files. Plus: a `node_modules/`-sized ignored entry is still skipped.

**Root cause.** Both checkpoints read `git status --porcelain`, which is blind to
`.gitignore`d entries — `task-deletion.ts:52` (the delete gate) and `salvage.ts:82` (salvage's
early return). The comment above that early return — *"Nothing uncommitted (tracked or
untracked) → nothing a force-remove could destroy"* — is false for ignored files, and
`smallIgnoredPaths` ten lines below it proves the author knew. `salvage-ignored.ts` was
correct code made unreachable by the check above it.

**Fix.** A new `GitWorktreeManager.ignoredWork()` runs the same `smallIgnoredPaths` rule the
snapshot uses, and both delete gates consult it. `isDirty` is deliberately NOT changed:
gitignored files survive a land and a sync, so folding them in would make every worktree
holding a `.env` read dirty to the sidebar and to `land`'s preflight — a different and wrong
claim. Salvage no longer returns early when the porcelain is empty but ignored work exists.

Because gate and snapshot share one rule, the delete refuses for **exactly** what a `--force`
retry then rescues.

**Proof — BEFORE (unfixed):**
```
--- git status --porcelain:
(end)
--- git status --porcelain --ignored:
!! .scratch/
!! HANDOFF.md
--- delete WITHOUT --force:
{"taskId":"01M1PBQQHXH6ZCM0EY2G938X43","queued":true,"status":"removed"}
--- dir: ls: .../ignsalv-7ab8516af77e/flounder: No such file or directory
--- salvage refs: (empty)
```

**AFTER:**
```
--- delete WITHOUT --force:
{"error":{"message":"task 01M1PC18ERMYDB32C6TPSYNKPQ worktree has gitignored work git status
 cannot see: .scratch/, HANDOFF.md","code":"DIRTY_WORKTREE", …}}
--- dir still there?  HANDOFF.md  README.md
--- delete WITH --force: {"status":"removed"}
--- salvage refs: refs/rove/salvage/ign-20260904T140921Z
--- ls-tree:  .gitignore  .scratch/notes.md  HANDOFF.md  README.md
```

**Size-ceiling regression (both directions, measured):**
- 80 MB `node_modules/` as the ONLY ignored entry → `delete` **succeeds** with no force, no
  ceremony. (`du -sk` = 81920 kb vs the 65536 kb ceiling.)
- 80 MB `node_modules/` + 4 kb `.scratch/` → forced snapshot contains `.scratch/notes.md` and
  **not** `node_modules/`.

Note on cost: the gate adds one `git status --porcelain -z --ignored` plus one `du -sk` per
top-level ignored entry to the non-force delete path. On a repo whose ignored tree is a
1.4 GB `node_modules/`, `du -sk` walks it — the same walk salvage already paid on the force
path. I did not cap it; if it proves slow in practice the number to change is
`MAX_IGNORED_ENTRY_KB`'s measurement strategy, not the rule.

---

## A2 — salvage reported success while the snapshot held a gitlink

**PASS criterion:** either the snapshot contains `vendor/child/subwork.txt`, or the
`SalvageRecord` carries an explicit incompleteness signal the audit line surfaces, naming
`vendor/child`. Silence is the failure.

**Fix.** Took option (b) — the snapshot genuinely cannot hold a submodule's working tree.
After `write-tree`, the tree is read back and its `160000` entries land in a new
`SalvageRecord.uncaptured`, which `recoveryText` renders before the recovery commands.

**Proof — AFTER (daemon.log):**
```
salvaged task 01M1PC1ZNT8C3RVCTQER3MAFPW — uncommitted work saved to
refs/rove/salvage/sub-task-20260904T140943Z (d80705d…). NOT captured (submodule / nested
worktree — the snapshot holds only a commit pointer): vendor/child. Recover with: …
```
BEFORE, the same run produced the ref and the `git restore` advice with no mention of
`vendor/child`; `git ls-tree -r <ref>` showed `160000 commit … vendor/child` and
`subwork.txt in snapshot: 0`.

---

## A3 — a nested worktree under a gitignored path destroyed the same way

**PASS criterion:** the no-force delete is refused, naming `.scratch/nested`; the control
(non-ignored `.scratch/`) stays green.

**No separate probe was needed** — and that is the finding. `git status --porcelain --ignored`
in the parent reports `!! .scratch/`, so A1's gate covers this shape too. I verified that
rather than assuming it. No nested-worktree special case was added, so no new ceremony
exists; the size budget gives the right answer on its own.

**Proof — BEFORE:** `{"status":"removed"}`, nested dir gone, `nested-precious.txt` destroyed,
salvage refs empty.
**AFTER:** refused with `gitignored work git status cannot see: .scratch/`, and
`cat .scratch/nested/nested-precious.txt` → `NESTED-PRECIOUS-UNCOMMITTED`.

---

## B — `removed` reported for a branch left permanently wedged

**PASS criterion:** after `delete --wait` reports success, `git worktree list` shows only the
main checkout, `git branch -D vanish2` succeeds, and `discover-adoptable` returns `[]`.

**Root cause.** `manager-remove.ts:176` ran git discovery from `path.dirname(worktreePath)` —
`~/.rove/worktrees/<key>`, which is inside no repository — so `findRepoFor` returned null and
the prune never ran. The owning repo was sitting in `task.repo` the whole time and was never
passed down.

**Fix.** `RemoveOpts.repo`, preferred over discovery in the missing-directory branch, threaded
from `TaskDeletionCoordinator.finish`. The discovery fallback stays for callers holding only a
path (the worktrees page), which is where it can still work. Ordering is unaffected: the
directory is already gone in this branch, so the prune touches admin records only.

**Proof — BEFORE:**
```
--- rev-parse from parent of worktree: fatal: not a git repository
--- delete: {"status":"removed"}
--- worktree list: …/gone2  2f26d4c [main]
                   /private/tmp/…/goat  05c91bb [vanish2] prunable
--- branch -D vanish2: error: cannot delete branch 'vanish2' used by worktree at …/goat
--- discover-adoptable: {"worktrees":[{…"dirty":null,"kobeManaged":true}],"unreadable":[]}
```
**AFTER:**
```
--- delete: {"status":"removed"}
--- worktree list: …/gone2  c4dea42 [main]        ← only the main checkout
--- branch -D vanish2: Deleted branch vanish2 (was b2178fb).
--- discover-adoptable: {"worktrees":[],"unreadable":[]}
```

---

## Screenshots — the TUI half of A1

The two-stage force confirm `docs/WORKTREES.md` calls "the boundary that authorizes data
loss" never fired for an ignored-only worktree. Captured through the only ground-truth path
(`/harness` browser → xterm.js → PTY sidecar → real OpenTUI), same key sequence both times,
on two separate fixtures with their own port bases:

- **BEFORE** (source reverted to pre-fix, rebuilt, fresh fixture on base 5883):
  `/Users/jacksonc/.rove/worktrees/kobe-0aff3858ab76/gemsbok/.scratch/loop-r15-ao/shots/before.png`
  One confirm and the task is gone — sidebar reads *"No active tasks — create one above."*
  On disk: worktree unlinked, `git for-each-ref refs/rove/salvage` empty.
- **AFTER** (base 5873):
  `/Users/jacksonc/.rove/worktrees/kobe-0aff3858ab76/gemsbok/.scratch/loop-r15-ao/shots/after.png`
  The second dialog fires: *"Visual Fixture" has uncommitted changes — "visual-fixture" has
  uncommitted, untracked, or gitignored work…* The task is still in the sidebar and
  `HANDOFF.md` still reads `A WHOLE SESSION OF REASONING`.

Both PNGs are attached as a comment on this PR. They live under `.scratch/`, which is
gitignored — i.e. in exactly the directory this bug destroys, which is why they could not be
committed to the branch.

---

## Tests

Four regression tests, all real git, no mocks. Each was mutation-verified — the fix reverted
at one site, exactly one test goes red:

| Mutation | Red test |
|---|---|
| drop the `ignoredWork` gate in `manager-remove.ts` | `an ignored-only worktree is refused without force…` |
| restore salvage's empty-porcelain early return | same test (the salvage half) |
| make `uncaptured` always `[]` | `a submodule's uncommitted work is reported as uncaptured…` |
| re-discover the repo from `dirname` | `prunes the stale admin record using the repo the caller passed` |

`an ignored tree over the size budget still deletes with no force` guards the other direction.

**Gates:** repo-root `bun run lint` clean · `bun run typecheck` clean · vitest **4495 passed** ·
`bun test test/render` **474 passed** · `KOBE_INCLUDE_SOCKET=1 test:socket` **807 passed** ·
`bun run check-i18n` — 834 keys × 2 locales in sync.

The full suite caught one thing I had gotten wrong: `smallIgnoredPaths` runs `git status`
through `exec.run` directly, bypassing `runGit`'s `READ_ONLY_GIT_ENV`. Harmless while it only
ran on the force path; now that it runs on every ordinary delete, it would contend with an
engine's `git commit` for `.git/index.lock`. `git-readonly-env.test.ts` failed and named it.
Fixed by passing `READ_ONLY_GIT_ENV` explicitly.

One test expectation was updated deliberately: `api-handlers-delete.test.ts` pinned the exact
recovery prompt, which now has to cover a path `git status` will not show.

## Copy and docs

- `DirtyWorktreeError` names the ignored paths — a user told only "uncommitted or untracked
  changes" would go looking with a command that reports nothing.
- The force-confirm body now reads "uncommitted, untracked, or gitignored work", both locales,
  `check-i18n` green.
- `docs/WORKTREES.md`: step 2 of the two-stage flow documents the `--ignored` check and the
  shared budget; the salvage section documents the submodule / nested-worktree gap.

## Not done / out of scope

- **C1, C2, D** left for another worker, as instructed.
- `deleteBranch: true` on a task whose directory is already gone still leaks the branch: the
  branch name is read via `currentBranch(worktreePath)`, which needs the directory. Same class
  of defect as B (a verb reporting success while doing nothing), but a different symptom, no
  repro in the brief, and it needs another field threaded — flagging rather than widening.
- No new or moved keybinding chords.

### DIFF
--- a/.changeset/dirty-gate-sees-gitignored-work.md
+++ b/.changeset/dirty-gate-sees-gitignored-work.md
@@ -1,0 +1,31 @@
+---
+"@sma1lboy/rove": patch
+---
+
+Deleting a task no longer destroys gitignored work without asking.
+
+The delete gate and the salvage snapshot both read `git status --porcelain`,
+which is blind to `.gitignore`d files — so a worktree whose only work was a
+`HANDOFF.md` or a `.scratch/` directory read as clean, deleted with no force
+and no confirmation, and wrote no salvage ref. Both now also ask
+`git status --ignored`, under the same 64 MB per-entry budget the snapshot
+already used: the delete refuses for exactly what a `--force` retry then
+rescues, and names the paths, because `git status` will not. A `node_modules/`
+is over that budget and still deletes with no ceremony. This also covers a
+nested worktree parked under a gitignored path, which was destroyed the same
+silent way.
+
+A salvage snapshot that could not capture everything now says so. `git add`
+records a submodule or nested worktree as a commit pointer rather than its
+files, so uncommitted work inside one was in neither the snapshot nor the
+commit that pointer named — while the audit line still told you to
+`git restore` from it. Those paths are now listed as `NOT captured`.
+
+Deleting a task whose worktree directory was already gone now actually
+deregisters it. The stale `.git/worktrees/` record can only be pruned from the
+owning repo, and the prune was looking for that repo by walking up from
+`~/.rove/worktrees/<key>` — a directory inside no repository — so it never
+ran. The delete reported `removed` while `git worktree list` still showed the
+entry as `prunable`, `git branch -D` failed forever with "used by worktree at
+<gone path>", and the worktrees page kept offering the ghost for adoption. The
+task's own repo is passed down now.
--- a/docs/WORKTREES.md
+++ b/docs/WORKTREES.md
@@ -141,6 +141,15 @@ The first confirmation can never silently turn into a force deletion. The
 second confirmation is the boundary that authorizes data loss. The branch is
 still retained, but uncommitted and untracked files are not part of it.
 
+Step 2 checks more than `git status --porcelain` reports. A gitignored
+`HANDOFF.md` or `.scratch/` shows in no status output, so the check also asks
+`git status --ignored` and refuses on any ignored entry under the 64 MB
+per-entry budget below — the same budget the snapshot uses, so the delete
+refuses for exactly what the forced retry then rescues. The refusal names
+those paths, because `git status` will not. An ignored entry OVER the budget
+(a `node_modules/`, a build directory) is not treated as work: it neither
+blocks the delete nor lands in the snapshot.
+
 ### Landing with `--delete-branch`
 
 `land --delete-branch` deletes the branch with `git branch -D`, which removes
@@ -182,6 +191,13 @@ build directory is far larger and is skipped, because a snapshot carrying one
 is too big to be useful. An ignored entry whose size cannot be read is
 skipped. So a gitignored `HANDOFF.md` or `.scratch/` is recoverable, and a
 gitignored 200 MB `dist/` is not.
+
+One thing a snapshot cannot hold: a submodule or a nested worktree. `git add`
+records those as a commit pointer rather than their files, so uncommitted work
+inside one is in neither the snapshot nor the commit that pointer names. Rove
+does not pretend otherwise — the audit line lists those paths as `NOT
+captured`, and the `git restore` commands below will not produce anything
+under them.
 
 List the snapshots, newest last:
 
--- a/packages/kobe-daemon/src/daemon/task-deletion-audit.ts
+++ b/packages/kobe-daemon/src/daemon/task-deletion-audit.ts
@@ -111,10 +111,19 @@ export function auditDeletionFailed(taskId: string, task: DaemonTask | undefined
  * lost work, so the commands ship with it. `repo` scopes them with `-C` when
  * known (a task carries its repo; a bare worktree path does not).
  */
-function recoveryText(ref: string, commit: string, repo?: string): string {
+function recoveryText(ref: string, commit: string, repo?: string, uncaptured: readonly string[] = []): string {
   const at = repo ? ` -C ${repo}` : ""
+  // A submodule or nested worktree is in the tree as a `160000` gitlink — a
+  // commit SHA, never the files — so `git restore --source` cannot produce
+  // anything under it. Naming those paths is the difference between advice
+  // that works and advice that fails silently on the one path the user cares
+  // about most.
+  const gap =
+    uncaptured.length > 0
+      ? ` NOT captured (submodule / nested worktree — the snapshot holds only a commit pointer): ${uncaptured.join(", ")}.`
+      : ""
   return (
-    `uncommitted work saved to ${ref} (${commit}). Recover with: ` +
+    `uncommitted work saved to ${ref} (${commit}).${gap} Recover with: ` +
     `git${at} show ${ref}  |  git${at} restore --source=${ref} -- <path>  |  ` +
     `list all: git${at} for-each-ref refs/rove/salvage`
   )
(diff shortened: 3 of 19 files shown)


Sma1lboy:
## Harness screenshots — the TUI half of A1

Captured through `/harness` → xterm.js → PTY sidecar → real OpenTUI. Same key sequence both
times (`click` the task row · `ctrl+q` back to sidebar · `d` · `right` · `enter`), on two
separate fixtures with their own port bases.

They are PNGs on the dispatcher's disk — `gh` cannot upload images, and they live under
`.scratch/`, which is gitignored, so they could not ride the branch either. That is the same
directory this bug destroys, which is the point.

**BEFORE** — source reverted to pre-fix, rebuilt, fresh fixture on base 5883:
`/Users/jacksonc/.rove/worktrees/kobe-0aff3858ab76/gemsbok/.scratch/loop-r15-ao/shots/before.png`

> One confirm and the task is gone. Sidebar reads *"No active tasks — create one above."*
> On disk: the worktree is unlinked and `git for-each-ref refs/rove/salvage` is empty, so
> `HANDOFF.md` and `.scratch/notes.md` are unrecoverable.

**AFTER** — fixed source, base 5873:
`/Users/jacksonc/.rove/worktrees/kobe-0aff3858ab76/gemsbok/.scratch/loop-r15-ao/shots/after.png`

> The second-stage confirm fires: *"Visual Fixture" has uncommitted changes — "visual-fixture"
> has uncommitted, untracked, or gitignored work. Force delete removes the worktree, but the
> work is snapshotted first to a salvage ref…* The task is still in the sidebar, and
> `HANDOFF.md` still reads `A WHOLE SESSION OF REASONING`.

Both states were confirmed on disk after the capture, not read off the pixels.

PR #885, opened 2026-09-04
After a reboot, a task's real conversations are reachable and named (B1, B2, B3)

Sma1lboy:
Batch B of `loop-r17-resume` (B1, B2, B3). One theme: after a pty-host restart
(a reboot, a crash, the host killed) every terminal tab is thawed-but-dead —
the frozen record keeps its command, cwd, geometry and scrollback, and the host
lists it. Three verbs told a headless caller otherwise, so the practical state
was: **after a reboot, every task's real conversation is unreachable until a
human opens the TUI.**

Nothing here auto-adopts or auto-resumes anything. B2 and B3 are pure
disclosure; B1 makes the respawn a TUI attach already performs *askable*, and
refuses by default.

All three were reproduced and re-measured in an isolated fixture (its own
`ROVE_/KOBE_ HOME_DIR / DAEMON_SOCKET_PATH / DAEMON_PID_PATH / PTY_SOCKET_PATH
/ PTY_PID_PATH / DAEMON_WEB_PORT=7811`, run from source, not the 0.9.105 on
PATH). Evidence transcripts: `/tmp/loop-r17-az-evidence/`.

Isolation verified, not assumed: `grep -c "scratch/loop-r17-az/home"` and a
grep for both fixture task ids against the real `~/.rove/tasks.json` are both
0; the fixture daemon and pty host are stopped.

---

## B2 — `send` names the conversations it passed over

**PASS criterion.** With every real conversation frozen, a `send` that starts a
fresh session must disclose the frozen tabs it did not use.

**BEFORE** (`/tmp/loop-r17-az-evidence/before-b2-b3.txt`) — tab-1/2/3 all
freeze-restored:

```json
{"ok":true,"session":"…::tab-1","started":true,"engineReady":true,"delivered":true}
```

`get-task` then reported `running: true`, because the blank tab is alive.

**AFTER** (`/tmp/loop-r17-az-evidence/after-b2.txt`), same fixture, same state:

```json
{
  "ok": true, "session": "…::tab-1", "started": true, "engineReady": true, "delivered": true,
  "frozenTabs": [
    { "tab": "tab-2", "sessionId": "dbd0931c-af9b-44ae-9c81-abeac22ced88" },
    { "tab": "tab-3", "sessionId": "83a6c407-8f8e-4533-8828-f4035270b692" }
  ]
}
```

**One correction to the brief.** It says "*No field distinguishes the two
branches*". `started: true` already did — its docstring is "A NEW session was
created by this call (never 'delivered into one')", and it was in the measured
payload. What was genuinely absent is the *context*: a first start of a healthy
task and "I opened a blank agent because your two real conversations are
frozen" both report `started/engineReady/delivered: true`. `frozenTabs` is that
missing half, and it names the tabs and the ids to reach them with. No
behaviour change — the branch was already chosen, this reports which one.

## B1 — a frozen tab is reachable, and the refusal is actionable

**PASS criterion (option b + a).** `send --tab tab-N` against a restored tab
must not be `TAB_NOT_FOUND` pointing at a listing that shows the tab; and the
respawn a TUI attach performs must be reachable headlessly.

**BEFORE** (`before-b1.txt`) — host back up, `pty-list` listing tab-2:

```json
{"error":{"message":"tab tab-2 has no live session on task … — see `rove api pty-list` for alive tabs","code":"TAB_NOT_FOUND"}}
```

**AFTER, default** (`after-b1-refusal.txt`) — no flag, nothing opened:

```json
{"error":{
  "message":"tab tab-2 on task … is a freeze-restored session — its pty host restarted, so the tab exists with its scrollback but nothing is running in it",
  "code":"TAB_RESTORED",
  "hint":"revive it with --respawn (resumes the tab's pinned conversation when it has one — `get-task` shows `sessionId`; a tab without one replays its recorded launch command)",
  "nextCommandArgs":["api","send","--task-id","…","--tab","tab-2","--respawn","--prompt","wake tab2"]
}}
```

**AFTER, with consent** (`after-b1-respawn.txt`):

```json
{"ok":true,"session":"…::tab-2","started":false,"delivered":true,"promptEcho":"confirmed","respawned":true}
```

### Why the respawn is explicit, and why the first-prompt hazard is gone

The brief flagged that an implicit respawn re-runs the recorded command, which
for a claude tab replays the task's original first prompt as an argv argument.
Two answers, both in this PR:

**1. The replay is avoided, not accepted.** The respawn composes its argv
through the same `engineTabArgv` a TUI dead-reattach uses, so a tab that
spawned before and has a pinned id comes back on `--resume <id>` with the
prompt *pasted* after the engine is up. Proof from the tab's own ring — the
original spawn and the respawn, same conversation:

```
SENTINEL claude UP pid=71843 argv=[--session-id dbd0931c-af9b-44ae-9c81-abeac22ced88 second conversation]
SENTINEL claude UP pid=91305 argv=[--resume dbd0931c-af9b-44ae-9c81-abeac22ced88]
```

**2. It is still explicit.** A tab with *no* pinned conversation id (a
non-resumable engine, a tab that never spawned) has no resume argv, so the host
falls back to the frozen launch command — the replay. And even a clean resume
starts a process and spends engine quota the caller did not ask for; `send
--tab tab-N` has never spawned anything. So the default stays a refusal and
`--respawn` is the consent, with the hazard spelled out in both the flag help
and the error hint. **New CLI flag — flagging it explicitly per the contract**
(`--respawn` on `send`, valid only with `--tab tab-N`; using it with
`--tab new`/canonical is a `BAD_FLAG` rather than a silent no-op).

## B3 — the pinned resume id is readable, and recorded correctly

**PASS criterion.** `get-task` returns `sessionId` per engine tab including
dead ones; and a headlessly-started tab has its id persisted.

**BEFORE** (`before-b2-b3.txt`) — every row: `id / kind / title / vendor /
liveVendor / lastTitle / autoTitle / alive / engineAlive / exit`, no id.

**AFTER** (`after-b3-read.txt`), `get-task` and `collect` alike:

```
{"id": "tab-1", "alive": false, "engineAlive": false, "sessionId": "c47b02fc-…"}
{"id": "tab-2", "alive": false, "engineAlive": false, "sessionId": "dbd0931c-…"}
{"id": "tab-3", "alive": false, "engineAlive": false, "sessionId": "83a6c407-…"}
```

### The write path was worse than "null for agent-started work"

The brief predicted the field would be null for a CLI-started tab. Measured, it
is **wrong**, not absent. `publishCliTabSnapshot` is write-once — it no-ops
when the task already has a snapshot — so a canonical `send` that respawns
tab-1 after a restart pins a *new* conversation while the snapshot still names
the previous one:

BEFORE (`before-b3-write.txt`):

```
state.json  tab-1 sessionId = c47b02fc-09a2-4c36-a051-d8de2ee5b43b
live argv   --session-id 17aad766-3c77-4f90-8ec9-5b849162ab99 'revive me'
```

`--resume` off that field would have reopened the wrong conversation. AFTER
(`after-b3-write.txt`), the two agree:

```
state.json  tab-1 sessionId = f478c27b-b962-483a-acd6-32c1d1bbfe36
live argv   --session-id f478c27b-b962-483a-acd6-32c1d1bbfe36 'revive me'
```

Recording the id of a session *this process just started* is the same write
`--tab new` already does; it does not fight a mounted TUI for tab state.

---

## Tests

Three new cases in `test/cli/pty-delivery.test.ts`, chosen as the tightest
proof of the branches the fixture exercised once: the `TAB_RESTORED` refusal
(and that it opens/writes nothing), the opt-in respawn, and the `frozenTabs`
disclosure. Mutation-verified — `if (session?.restored === true)` → `if
(false)` and dropping `...disclose` from the success return fails exactly those
three and nothing else.

## Gates

- `bun run lint` — 0
- `bun run typecheck` — 0
- `bun run test:fast` — 451 files, 4489 tests, 0 fail
- `bun test test/render` — 108 files, 477 tests, 0 fail
- `KOBE_INCLUDE_SOCKET=1 bun run test:socket` — 100 files, 819 tests, 0 fail

## Notes for review

- **No harness screenshots.** These three items change API/CLI response shapes
  only; nothing renders differently in the TUI, so a `/harness` capture would
  show two identical frames. The verb transcripts above are the ground truth
  the task's own proof standard asks for ("PASS is the verb telling the
  truth"). Saying so rather than attaching a decorative pair.
- **`sessionId` is joined in `runtime.ts`, not in `joinTaskTabs`.** The tidier
  home is `TaskTabRow` itself, but `tab-snapshot.ts` is owned by the Batch A
  worker this round (A3 edits the same object literal). `runtime.ts` already
  holds the snapshot, so the join is three lines there and folding it into
  `joinTaskTabs` is a one-line follow-up once A lands. Noted in the type's
  docstring too.
- **Two files split** to stay under the 500-line cap, both along stated seams:
  `exact-tab-delivery.ts` (deliver into ONE addressed tab, including reviving
  it) out of `pty-delivery.ts` (find or create the task's engine session); and
  `handlers-lifecycle.ts` (delete / land / adopt — the verbs that END a task
  and each own a half-teardown recovery story) out of `handlers-tasks.ts`,
  mirroring the existing `verbs-lifecycle.ts` grouping.
- **Docs updated in the same PR** per AGENTS.md: `docs/API.md` (`get-task`
  `sessionId`, `send --respawn` / `TAB_RESTORED` / `frozenTabs`) and
  `docs/SESSIONS.md` ("Resuming a conversation" no longer sends the reader to
  claude's `/resume` picker for an id Rove can print). `api-flags.md`
  regenerated with `bun scripts/gen-skill-api-flags.ts`.
- **Not touched**, per the task's boundaries: `cwd-task.ts`,
  `handlers-engine-report.ts`, `tab-snapshot.ts`, the worktree manager,
  `refs/`, `.github/workflows/*`.

### DIFF
--- a/.agents/skills/kobe/references/api-flags.md
+++ b/.agents/skills/kobe/references/api-flags.md
@@ -98,7 +98,7 @@ needs numbers, not when you want to know what a task is doing.
 
 <!-- generated:begin drive -->
 ```text
-send              --task-id --prompt|--prompt-file(REQ) --tab --command --plain
+send              --task-id --prompt|--prompt-file(REQ) --tab --command --respawn --plain
                   --allow-empty
 dispatch          --task-id(REQ) --prompt|--prompt-file(REQ) --tab
 deferred-list     --task-id
--- a/.changeset/frozen-tabs-reachable.md
+++ b/.changeset/frozen-tabs-reachable.md
@@ -1,0 +1,26 @@
+---
+"@sma1lboy/rove": patch
+---
+
+After a pty-host restart, a task's real conversations are reachable and named
+
+A reboot (or any pty-host restart) freezes every terminal tab: the session
+record keeps its command, cwd and scrollback, and the host lists it. Three
+verbs told a headless caller otherwise.
+
+- `send --tab tab-N` refused a frozen tab with `TAB_NOT_FOUND` and pointed at
+  `pty-list`, which lists that very tab. It now refuses with `TAB_RESTORED`
+  and a hint, and `send --respawn` revives the tab in place and delivers into
+  it — resuming its pinned conversation (`--resume <id>`) rather than
+  replaying the task's first prompt. The respawn is never implicit: a tab
+  with no pinned id would re-run its recorded launch command.
+- `send` with no `--tab` fell through to a never-started tab, spawned a blank
+  session and reported plain success while both real conversations sat
+  frozen. It now lists them as `frozenTabs` (`tab` + `sessionId`) whenever it
+  started a new session.
+- `get-task` / `collect` now return each engine tab's `sessionId` — the id
+  the engine's resume verb needs, persisted since engine tabs existed and
+  exposed nowhere. The canonical `send` path also records the id of a session
+  it just started, which the write-once snapshot seeding skipped: a tab
+  respawned after a restart was pinned to a new conversation while the
+  snapshot still named the previous one.
--- a/claude-plugin/skills/rove/references/api-flags.md
+++ b/claude-plugin/skills/rove/references/api-flags.md
@@ -98,7 +98,7 @@ needs numbers, not when you want to know what a task is doing.
 
 <!-- generated:begin drive -->
 ```text
-send              --task-id --prompt|--prompt-file(REQ) --tab --command --plain
+send              --task-id --prompt|--prompt-file(REQ) --tab --command --respawn --plain
                   --allow-empty
 dispatch          --task-id(REQ) --prompt|--prompt-file(REQ) --tab
 deferred-list     --task-id
--- a/docs/API.md
+++ b/docs/API.md
@@ -137,7 +137,12 @@ replacement in `nextCommandArgs`.
 - `get-task --task-id <id>`: one task's metadata; `.running` = any of its
   hosted engine tabs is live (not just the first); `.tabs` = the task's
   terminal tabs (`id`/`kind`/`title`/`vendor`/`liveVendor`/`lastTitle`/
-  `autoTitle` + per-tab `alive`): the discovery read for `send --tab tab-N`.
+  `autoTitle`/`sessionId` + per-tab `alive`): the discovery read for
+  `send --tab tab-N`. `sessionId` is the conversation the tab pinned — the
+  exact id `claude --resume <id>` (or the engine's own resume verb) reopens,
+  and the one `send --tab tab-N --respawn` brings back. Present on engine
+  tabs that recorded one, dead tabs included; absent for engines that mint
+  their own id late and for tabs that never spawned.
   `liveVendor` on a live tab is a fresh foreground process walk (which
   engine actually runs in the tab's shell right now. A hand-typed `claude`
   in a shell tab counts, a ctrl+C'd engine doesn't); dead tabs report the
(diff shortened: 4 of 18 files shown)


Sma1lboy:
Follow-up on the `file-size-cap` red: it was `test/cli/pty-delivery.test.ts` (451 on main → 523 here), and there **was** a seam — the one this PR's own source split already drew.

`deliverToExactTab` moved to a new `test/cli/exact-tab-delivery.test.ts`, mirroring `src/cli/api/exact-tab-delivery.ts`. Boundary: **`pty-delivery.test.ts` owns delivery through the canonical path** (find the task's engine session, or create it — including the new `frozenTabs` disclosure, which is a `deliverHostedPrompt` fact); **`exact-tab-delivery.test.ts` owns delivery when the caller named the tab**, live or gone (absent, a plain shell, or thawed-but-dead after a host restart). `pty-delivery-gates.test.ts` keeps the composer-busy gates on the same function, unchanged. 523 → 438 + 129, and imports run test → src only, no shared module and no cycle.

Also corrected a wording inaccuracy I introduced, in the same push. The `--respawn` help, the `TAB_RESTORED` hint and both doc pages said a tab with no pinned id "replays its recorded launch command". Reading my own code again, that is only true for a tab Rove holds **no snapshot** for: when the snapshot has the tab, the launch is composed fresh with `promptIntent: \"none\"`, so a tab without a `sessionId` comes back on a fresh conversation with no prompt in its argv. The replay case is narrower than I first wrote, and it is still the reason the flag is explicit. All three surfaces now say that.

Re-verified locally: root `lint` 0, `typecheck` 0, `test:fast` 454 files / 4496 tests / 0 fail. Every touched `.ts` is now ≤438 lines, so no exemption is needed.

PR #888, opened 2026-09-04
fix: close the i18n gate's reachability hole, and stop a Chinese label being cut with no ellipsis

Sma1lboy:
Three batches from `.scratch/loop-r18-i18n/i18n.md`, in the explorer's order (3 → 1 → 2). One theme: **`check-i18n` validates PARITY, so a key missing from BOTH locales passes it** — and everything below is either that hole or a string that never became a key at all.

Every screen-visible claim is measured off the **terminal text buffer** through `/harness` → xterm.js → PTY → real OpenTUI, same chord, same fixture, only `state.json.locale` changed. BEFORE frames were captured on base code (`d542d1a30`, `git checkout <base> -- packages/kobe/src` + rebuild, then restored); AFTER on the built branch. Port base 7311, isolated home under `.scratch/opentui-visual-7311/`, fixture and its PTY host both torn down and verified gone.

Evidence (absolute paths, outside `.scratch/`): `/tmp/loop-r18-ba/`

---

## Batch 3 — a cell budget spent by a code-point truncator

`prefix-hud.tsx` fed `width - 2` (CELLS) into `truncateEnd` (CODE POINTS). A Chinese caption is 25 points against a 26 budget, so it "fit", came back whole, and `wrapMode="none"` let Yoga shear it — landing right after an **opening** full-width bracket, with no ellipsis to say anything was dropped.

**PASS criterion:** the zh HUD line ends in `…` and fits the same cell budget English does; English is unchanged.

```
BEFORE  zh   ctrl+a + 2 → 打开例行任务（      ← hard cut, no ellipsis
AFTER   zh   ctrl+a + 2 → 打开例行任务…
BEFORE  en   ctrl+a + 2 → Open routine…      ← control, unchanged
AFTER   en   ctrl+a + 2 → Open routine…
```

`/tmp/loop-r18-ba/{before,after}-hud-zh.png` · `{before,after}-hud-zh.txt` · `{before,after}-hud-en.txt`

Fixed at the same time: `truncateTitle`, whose `max` every caller measures with `approxCellWidth` while it spent it in code points. That is the context-menu call site from 3.2 — one function, both callers, not a per-caller patch. Its doc already claimed "cell budget"; now it is one.

Regression tests: `test/render/which-key-help.test.tsx` renders the real `PrefixHud` at width 28 in zh and asserts the row contains `打开例行任务…`, does **not** contain `定时任务）`, and stays inside the budget. **Mutation-verified** — reverting only the call site to `truncateEnd`:

```
Expected to contain: "打开例行任务…"
Received: "  ctrl+a + 2 → 打开例行任务（   …"     ← byte-identical to the harness BEFORE
```

## Batch 1 — a test that documented its own gap, delegating to a guard that did not exist

`workspace.reopenSession` had no `keys.desc` entry in **either** locale. `tKeys()` falls back to the raw lookup key — which for this group is the binding id — so the F1 help dialog printed the literal string `workspace.reopenSession` as that row's description. The correct English sentence has been sitting unread at `keybindings-chat.ts:146` the whole time; the dialog never reads the binding's `description` field. `Diff review` and `Inbox` had the same gap in `keys.category`.

**PASS criterion:** no binding id or category header renders raw, and deleting a `keys.desc` entry from both locales — which keeps `check-i18n` green — fails CI.

```
BEFORE zh  ⏎   workspace.reopenSession                       (⏎)
AFTER  zh  ⏎   在标签页全部关闭的任务中重新打开会话            (⏎)
```
`/tmp/loop-r18-ba/{before,after}-help-zh.png` · `{before,after}-help-zh.txt`. Grep for `reopenSession` across the AFTER dialog: **0 occurrences**.

`test/tui/i18n-key-resolution.test.ts` now covers what its own module doc previously carved out by name (paragraph deleted). Two tests, driven from the keymap and the two category mappers rather than from source text: every binding id resolves in `keys.desc`; every category header any surface can print resolves in `keys.category` — unioning the help dialog's scope mapping, the prefix HUD's guide mapping, and the keymap's own `category` field, as the brief specified.

**Mutation-verified at two independent sites**, each showing the exact shape the guard exists for:

| mutation | `check-i18n` | new guard |
|---|---|---|
| delete `keys.desc["sidebar.rename"]` from **both** locales | `OK — 842 keys × 2 locales in sync` | `expected [ 'sidebar.rename' ] to deeply equal []` |
| delete `keys.category["Diff review"]` from **both** locales | `OK — 842 keys × 2 locales in sync` | `expected [ 'Diff review' ] to deeply equal []` |

To make that possible, `scopeCategory` (help-dialog.tsx) and `guideCategory` (prefix-hud.tsx) moved to the framework-free `lib/help-groups.ts` — a vitest guard cannot import an opentui component to ask which header a binding prints under, and duplicating their literals into the test would guard the copy rather than the code.

## Batch 2 — ~30 un-catalogued strings, fixed at the mechanism in each owner

Not a re-translation pass: each owner got the machinery it was missing, and the wording followed.

**The Routines schedule preview** — the one line answering "when does this fire", and the last fully-English line in a Chinese composer.

```
BEFORE  weekdays at 09:00 · in 2d · Mon 09:00
AFTER   工作日09:00 · 2 天后 · 周一 09:00
```
`/tmp/loop-r18-ba/{before,after}-composer-zh.png` · `{before,after}-composer-zh.txt` (the BEFORE frame carries both this line and Batch 3's clipped HUD).

- `describeCron` / `describeTimeOfDay` read their phrases from the catalog. `describeTimeOfDay` now **reports** whether its phrase is already self-recurring instead of sniffing an `"every "` prefix — English could detect that, a translated phrase cannot, and without it zh would read "每天每 15 分钟".
- `formatRelative` / `formatWhen` likewise.
- `formatAbsolute` takes its calendar words **and their order** from `Intl` for the active locale, deleting the `WEEKDAYS`/`MONTHS` tables. zh reads `2026/8/3`, not `8/3/2026`; the explorer's "give zh its own template" is unnecessary once the platform owns word order. The 24-hour clock is still built by hand — same instant everywhere, and `Intl`'s en-US default would turn it into `9:00 AM`.

**Two `toLocale*` calls passed no locale at all**, following the OS instead of the UI setting — wrong in both directions. `intlLocale()` in `tui/i18n` is now the single owner of the BCP-47 tag, carried on `LOCALES` beside each catalog id.

**The daemon's run-status enum** is mapped through the catalog; an unmapped status falls through to its raw value so a new one is loud, not blank. `run.error` stays untranslated deliberately — it is a machine message the user searches for and pastes into an issue, and a translated copy matches nothing. (Written down here because the next reader will ask.)

**The Settings dialog's modals** — `Reset UI state?`, `Restart backend?`, `Add engine`, `Can't use that worktree location`, the `COMMAND`/`NAME`/`ID`/`PROTOCOL` field labels and `save`/`next`/`add` captions, and the two lines printed to stderr *after* Rove exits.

**The duplicate `relativeAge`** in `work-items-page.tsx` is gone in favour of the shared clock. This one changed real behavior and one test caught it: the local copy **rounded** where `src/lib/relative-time.ts` documents that every step must floor, and it had no seconds step, so a freshly-stamped row read `0m` where every other age on screen reads `3s`. `test/render/failure-toast.test.tsx` asserted that `0m`; it now asserts the age by shape (`/\b\d+[smhd]\b/`), since the fixture stamps `updatedAt` at module load and the literal depended on how long the file took to run.

Tests: `test/tui/automation-composer.test.ts` asserts the whole assembled zh preview line matches no `[A-Za-z]` (the explorer's stated criterion) and that a beyond-the-week date contains `2月1日`; `test/tui-react/automations-format.test.ts` covers `formatWhen` in both locales including the `>24h` Intl fallback, plus `formatRunStatus`'s fall-through. The resolution gate was extended to the two new enumerable dynamic key families (`automations.schedule.dow.<CODE>`, `automations.runStatus.<status>`), the latter driven from the daemon's union via an exhaustive `Record`, so adding a status there without a catalog entry is a **type** error.

---

## Gates

| gate | result |
|---|---|
| `bun run lint` (repo root) | clean, 1449 files |
| `bun run typecheck` | clean |
| `bun run test:fast` | 455 files, **4509 passed** |
| `bun test test/render` | 108 files, **479 passed** |
| `KOBE_INCLUDE_SOCKET=1 bun run test:socket` | 100 files, **819 passed** |
| `bun run check-i18n` | `900 keys × 2 locales in sync` (was 840) |

Rebased on `origin/main` @ `e91042fb6` (0.9.139); lint / typecheck / check-i18n / test:fast / render all re-run green after the rebase.

## Not done / notes

- **No new or moved chords.** Nothing in these batches needed one.
- **Batch 4 was read and acted on in exactly one place:** the explorer's correction that the broken-surrogate class is closed at the helper level is why Batch 3 is a call-site unit fix and not a helper rewrite. The catalog width audit was not re-derived.
- `describeTimeOfDay`'s return type changed from `string | null` to `{ at, bare } | null`. It is module-private; the 18 existing English assertions in `test/tui/cron-segments.test.ts` all still pass unchanged, which is the behavior-preserving proof.
- **Wording is a first pass.** The zh strings are mine and read correctly, but the mechanism is the deliverable — every one of them is now a catalog key that can be edited without touching code.

### DIFF
--- a/.changeset/i18n-gate-reachability.md
+++ b/.changeset/i18n-gate-reachability.md
@@ -1,0 +1,39 @@
+---
+"@sma1lboy/rove": patch
+---
+
+Close the hole the i18n gate's own comment described, and stop a Chinese label
+from being cut with no ellipsis.
+
+`check-i18n` validates PARITY — that `en` and `zh` agree with each other — so a
+key missing from BOTH locales passes it. `workspace.reopenSession` had no
+`keys.desc` entry in either, and because `tKeys()` falls back to the raw lookup
+key, the F1 help dialog printed the literal string `workspace.reopenSession` as
+that binding's description, in both languages. The binding's real English
+sentence has been sitting unused in the keymap table the whole time; the help
+dialog never reads it. Two keymap categories (`Diff review`, `Inbox`) had the
+same gap, unreached only by accident of which surfaces are wired today.
+
+`test/tui/i18n-key-resolution.test.ts` now covers what it previously carved out
+by name: every binding id must resolve in `keys.desc`, and every category header
+any surface can print — the help dialog's scope mapping, the prefix HUD's guide
+mapping, and the keymap's own `category` field — must resolve in `keys.category`.
+Deleting a `keys.desc` entry from both locales keeps `check-i18n` green and now
+fails this test, which is the shape of failure it exists to catch. The two
+category mappers moved to the framework-free `lib/help-groups.ts` so the guard
+can ask them directly instead of duplicating their literals.
+
+The prefix HUD's stroke echo fed a CELL budget into a code-point truncator, so a
+Chinese caption "fit" and Yoga then sheared it: `ctrl+a + 2 → 打开例行任务（`,
+ending on an opening full-width bracket with nothing to say the rest was
+dropped, where English got a clean `Open routine…`. Same mismatch in
+`truncateTitle`, whose budget every caller measures with `approxCellWidth`.
+
+The Routines schedule preview, its relative clocks, and the modals the Settings
+screen opens are no longer English-only inside a translated UI. Two
+`toLocale*` calls passed no locale at all and followed the OS rather than the UI
+setting — wrong in both directions — and the next-run date built its word order
+from an English weekday/month table; it now comes from `Intl` for the active
+locale, so zh reads `2026/8/3` rather than `8/3/2026`. The Work Items page's
+private copy of `relativeAge` (which rounded where every other age on screen
+floors) is gone in favour of the shared clock.
--- a/packages/kobe/src/tui-react/component/automations-format.ts
+++ b/packages/kobe/src/tui-react/component/automations-format.ts
@@ -7,17 +7,28 @@
  */
 
 import { relativeBuckets } from "../../lib/relative-time"
+import { intlLocale, t } from "../../tui/i18n"
 
 /** `in 5m` / `12m ago` / a date once it is more than a day out. `\u2014` when
- *  the timestamp is missing or unparsable — an absence, never a fake "now". */
+ *  the timestamp is missing or unparsable — an absence, never a fake "now".
+ *  The date falls back on `Intl` for the UI locale, not the machine's. */
 export function formatWhen(iso: string | undefined, now: number): string {
   if (!iso) return "—"
   const at = Date.parse(iso)
   if (!Number.isFinite(at)) return "—"
   const deltaMs = at - now
+  const ahead = deltaMs >= 0
   const { minutes, hours } = relativeBuckets(Math.abs(deltaMs))
-  if (minutes < 1) return deltaMs >= 0 ? "now" : "just now"
-  if (minutes < 60) return deltaMs >= 0 ? `in ${minutes}m` : `${minutes}m ago`
-  if (hours < 24) return deltaMs >= 0 ? `in ${hours}h` : `${hours}h ago`
-  return new Date(at).toLocaleDateString()
+  if (minutes < 1) return t(ahead ? "automations.when.now" : "automations.when.justNow")
+  if (minutes < 60) return t(ahead ? "automations.when.inMinutes" : "automations.when.minutesAgo", { n: minutes })
+  if (hours < 24) return t(ahead ? "automations.when.inHours" : "automations.when.hoursAgo", { n: hours })
+  return new Date(at).toLocaleDateString(intlLocale())
+}
+
+/** The daemon's run-status enum as display text; an unmapped status falls
+ *  through to its raw value so a new one is loud rather than blank. */
+export function formatRunStatus(status: string, translate: (key: string) => string): string {
+  const key = `automations.runStatus.${status}`
+  const label = translate(key)
+  return label === key ? status : label
 }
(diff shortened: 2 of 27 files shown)


Sma1lboy:
### Harness frames

All six captured through `/harness` → xterm.js → PTY → real OpenTUI at port base 7311, zh locale. BEFORE frames on base code `d542d1a30`, AFTER on this branch.

Local paths (the text-buffer dumps beside each are the actual assertion; the PNGs are the same frames):

| | BEFORE | AFTER |
|---|---|---|
| prefix HUD (Batch 3) | `/tmp/loop-r18-ba/before-hud-zh.png` | `/tmp/loop-r18-ba/after-hud-zh.png` |
| F1 help dialog (Batch 1) | `/tmp/loop-r18-ba/before-help-zh.png` | `/tmp/loop-r18-ba/after-help-zh.png` |
| Routine composer (Batch 2) | `/tmp/loop-r18-ba/before-composer-zh.png` | `/tmp/loop-r18-ba/after-composer-zh.png` |

The BEFORE composer frame is worth one look on its own: it carries **both** remaining defects in a single screen — the English `weekdays at 09:00 · in 2d · Mon 09:00` preview line, and the HUD's `打开例行任务（` clipped at the bottom-left.

PR #895, opened 2026-09-04
A pane cannot say which task it is in, and goes blind when the daemon restarts (Group 2 + 3)

Sma1lboy:
Loop r20 plugin exploration, **Group 3** (panes are second-class citizens of the documented contract) plus **Group 2** (a plugin cannot see the host it is talking to). Groups 1 and 4 landed separately in #894; nothing here touches `plugins/runtime.ts`'s hook execution or the registry poll.

Every claim below was measured from an author's seat: a real plugin (`bi.probe` / `bi.board`) linked into an isolated from-source daemon (0.9.142), plus the `/harness` → xterm.js → PTY sidecar → real OpenTUI path for the screen-visible ones. Evidence lives at **`/Users/jacksonc/rove-evidence/loop-r20-bi/`** (probe plugins, raw terminal buffers, screenshots).

---

## 1. A pane can name the task it is in (3.1)

**PASS:** a pane prints its resolved task id where it printed `(none)`.

`buildPaneArgv` injected only `ROVE_PLUGIN_ENTRYPOINT_ID`, and `plugin pane open --task <id>` resolved a task id one frame before composing argv and then threw it away. The only workaround left — matching `cwd` against `worktreePath` over `task.snapshot` — cannot be correct for project-main and directory tasks, which reuse an existing checkout and share one cwd.

Both openers now resolve the task first and pass `taskId` into `buildPaneArgv`, which injects `ROVE_PLUGIN_TASK_ID` (+ its `KOBE_` alias). No `_TASK_TITLE`: the TUI picker has no title to give, and a half-present variable is the same defect in a new place — the doc points at `api get-task --task-id "$ROVE_PLUGIN_TASK_ID"` instead.

**Proof** — the login-shell script the host hands a TUI, tapped off the daemon's `tab.open` channel (`{before,after}-pane-argv.txt`):

```
BEFORE:  (no PLUGIN_TASK_* variable in the env prefix at all)
AFTER:   ROVE_PLUGIN_TASK_ID=01M1PWXAXTSNGYA1SFEPC5SFW1
         KOBE_PLUGIN_TASK_ID=01M1PWXAXTSNGYA1SFEPC5SFW1
```

Running that exact argv under a real PTY and reading the pane's own text buffer (`{before,after}-pane-raw.txt`):

```
BEFORE:  BI BOARD   taskId=(none)                       daemon=connected  tasks=2
AFTER:   BI BOARD   taskId=01M1PWXAXTSNGYA1SFEPC5SFW1   daemon=connected  tasks=2
```

## 2. A pane is told when the daemon dies — including under `kill -9` (3.2)

**PASS:** a subscriber receives something actionable when the daemon goes away, on both paths.

`RoveSocket`'s close handler failed *pending requests only*. A subscriber holds no pending request, so it heard nothing — and because the hosted PTY survives a daemon restart by design, the pane kept drawing its last frame and looked live. `onClose(handler)` now fires once from `failAll`, and is deliberately silent for the plugin's own `close()`.

**Proof — real daemon, real `kill -9`** (not a mocked close event; the whole finding is that the real one is never delivered). `{before,after}-socket-{graceful,kill9}.txt`:

```
BEFORE  kill -9:   … event task.snapshot            ← last line, forever; probe still alive
AFTER   kill -9:   … event task.snapshot
                   … CLOSE-SIGNAL: daemon socket closed

BEFORE  graceful:  … event daemon.stopping          ← last line, forever
AFTER   graceful:  … event daemon.stopping
                   … CLOSE-SIGNAL: daemon socket closed
```

**Harness BEFORE/AFTER** (`/harness` → xterm → PTY → real OpenTUI, fixture on port base 7727; identical key sequence, fixture daemon `kill -9`'d ~10s before each capture):

| | |
|---|---|
| `BEFORE-daemon-killed.png` | `task: (none)` · `host: connected` — the frozen board, still claiming a host that is dead |
| `AFTER-daemon-killed.png` | `task: 01M1PXKAHH2WNA2KC9NCH7FDZG` · `host: UNREACHABLE — reconnecting` |

`after-1-board-live.png` is the same board before the kill, rendering the live task list.

*(The probe board latches on loss rather than reconnecting, so the captured frame is stable; the documented recipe reconnects.)*

## 3. `openPane()` can target a task and reports what actually happened (3.4)

**PASS:** JSON carrying `clients`, and `openPane()` accepts a task id.

`plugin pane open` printed prose and exited 0 whether or not any UI performed the split; its sibling `api pane-open` returns `clients` for exactly that reason. Copied the sibling.

```
BEFORE:  opened pane bi.probe.board in task 01M1PWXAXTSNGYA1SFEPC5SFW1
AFTER:   {"ok":true,"clients":2,"pane":"bi.probe.board","taskId":"01M1PWXAXTSNGYA1SFEPC5SFW1","title":"BI Board"}
```

SDK `openPane(id, { taskId })` appends `--task` and surfaces `clients` (falling back gracefully on an older host that still prints prose).

## 4. A plugin can ask which host it is talking to (2.1, 2.2)

**PASS:** the answer has one obvious spelling and both pages name it.

`hello` already returned `kobeVersion` + `capabilities` — the running daemon's real channel list — callable from a plugin today and documented on neither page, while PLUGIN-SDK.md told authors to "check the SDK version", which cannot answer the question. Added `RoveSocket.hello(): DaemonInfo`, which surfaces the build version under both `roveVersion` and `kobeVersion` (the wire field keeps its spelling) alongside `capabilities` / `protocolVersion` / `homeDir`.

**Proof** — from inside a plugin action against the from-source daemon (`after-hello.txt`):

```
$ ./r.sh plugin action invoke bi.probe.info
{"version":"0.9.142","caps":19}
```

0.9.142 is the *daemon*, not the 0.9.105 on `PATH` — which is 2.2's point, now stated in the `ROVE_BIN_PATH` row: in a dev checkout your callbacks run a different build than the daemon that launched you, and comparing `$ROVE_BIN_PATH --version` against `hello`'s version is how you notice.

I did **not** re-derive the SDK-vs-daemon channel drift; the report measured zero and this is about discoverability.

## 5. "`command` is always argv: never a shell" (3.3)

**I treated the BEHAVIOUR as the promise and fixed the doc.** The login shell is a deliberate choice cross-referenced to `session-launch.ts`, and silently changing how existing panes launch would break every pane whose `command[0]` resolves only on the user's rc-file PATH. The sentence now carries its pane exception and names the three consequences (rc-file PATH, `command[0]` resolution, rc output arriving before your first draw).

On the drawing hazard: `Pane.start()` already enters the alternate screen and clears it, so rc chatter stays on the primary screen. That is now stated in the Pane-kit section, together with "clear the screen yourself if you draw without the kit" — the honest shape, rather than implying panes are unprotected.

---

## Tests

Four mutations, each caught by exactly one test:

| Mutation | Test that failed |
|---|---|
| drop the task-var injection in `buildPaneArgv` | `pane-command > injects the task id…` |
| revert `failAll` to pending-only (the original bug) | `onClose > tells a subscriber the connection died…` |
| drop the deliberate-close guard | `onClose > stays quiet for the plugin's own close()` |
| `hello()` drops the `roveVersion` alias | `hello > reports the RUNNING daemon's build…` |

Gates: repo-root `bun run lint`, `bun run typecheck`, `test:fast` (457 files / 4521 tests), `bun test test/render` (479), `KOBE_INCLUDE_SOCKET=1 test:socket` (100 files / 829) — all green.

coverage-exemption: packages/kobe/src/tui-react/workspace/TerminalTabs.tsx — one line (`taskId: props.taskId` into the dialogs hook). The file reads 33.8% because it cannot be mounted under the render harness (baseline, unchanged by this PR); the behavior it enables is proven by the harness screenshots above.
coverage-exemption: packages/kobe/src/tui-react/workspace/use-tab-dialogs.ts — reads 0% on the VITEST track because vitest never imports it; it is exercised by the render track (`test/render/new-chat-flow.test.tsx`, updated here). Its react dependency is type-only, so `isRenderTrackOnly` classifies it into the wrong track.

## Not proven

Nothing in scope is unproven. Two notes on the setup, both artifacts of this checkout rather than of Rove:

- The harness probe connects through a short symlink to the daemon socket: this worktree's fixture home makes the socket path 109 bytes, past the 104-byte `sun_path` limit. Product behavior is unaffected — the CLI-path proofs used the real path directly.
- Fixture daemon and PTY host were both stopped after capture; `lsof` over my socket paths shows no leftovers.

### DIFF
--- a/.changeset/plugin-pane-task-id-and-socket-close.md
+++ b/.changeset/plugin-pane-task-id-and-socket-close.md
@@ -1,0 +1,11 @@
+---
+"@sma1lboy/rove": patch
+---
+
+A plugin pane can now name the task it runs in, and a plugin knows when the daemon goes away.
+
+Panes get `ROVE_PLUGIN_TASK_ID` (both openers: `plugin pane open` and the ctrl+e picker), so a pane no longer has to guess its task by matching its cwd against `worktreePath` — a match that is ambiguous for project-main and directory tasks, which share one cwd. `plugin pane open` takes `--task <id>` and prints the same `{ok, clients, title, taskId}` JSON as `api pane-open`, so a caller can tell "no attached UI performed the split" (`clients: 0`) from "opened"; exit 0 alone never said that.
+
+The SDK gains `RoveSocket.onClose(handler)`, called once when the connection dies. A subscriber holds no pending request, so the old socket — which failed only pending requests — told it nothing at all when the daemon crashed, and a hosted pane's PTY outlives the daemon, so the pane kept drawing a frame that looked live. `RoveSocket.hello()` returns the RUNNING daemon's build version and channel list, which is the only way to tell "this host is too old for that channel" from "nothing has happened yet". `openPane()` takes a `taskId` and surfaces `clients`.
+
+PLUGIN-AUTHORING.md's "`command` is always argv: never a shell" now carries its pane exception — panes do run through the user's interactive login shell — and PLUGIN-SDK.md's pane example handles a lost connection instead of modelling the frozen board.
--- a/docs/PLUGIN-AUTHORING.md
+++ b/docs/PLUGIN-AUTHORING.md
@@ -202,8 +202,17 @@ any = ["ctrl-c to interrupt"]    # at least one must appear
 # bottom_lines = 12              # trailing non-empty lines examined (default 12)
 ```
 
-`command` is always argv: never a shell, no expansion (panes expand only
-`$ROVE_PLUGIN_ROOT`). Unknown event names are warnings (forward compat);
+`command` is argv, never a shell, for events, startup, shutdown and
+actions: no expansion, no pipes, no globs. **Panes are the exception** — a
+pane runs through the user's interactive login shell (`sh -ilc`, the same
+launch path as an engine tab), expands `$ROVE_PLUGIN_ROOT`, and therefore
+inherits the rc-file environment: your `command[0]` must resolve on the PATH
+the user's `.zshrc`/`.bashrc` builds, not the daemon's, and anything those
+files print (version-manager chatter, MOTDs, banners) reaches the terminal
+before your first draw. The pane kit's `start()` handles that for you — it
+enters the alternate screen and clears it, so rc output stays on the primary
+screen; a pane that does not use the kit should clear the screen itself
+before its first frame. Unknown event names are warnings (forward compat);
 invalid types/patterns are install-time errors.
 
 `first_message_delivery` says how the CLI takes a session's first message.
@@ -287,7 +296,7 @@ Every plugin command gets, on top of the user's environment:
 
 | Variable | Meaning |
 |---|---|
-| `ROVE_BIN_PATH` | exec this to call back into Rove — the absolute path of the running install when that is a runnable file (an npm install, a compiled binary), otherwise the bare `rove`/`kobe` name resolved on `PATH`, which is what a dev checkout run through `bun` falls back to |
+| `ROVE_BIN_PATH` | exec this to call back into Rove — the absolute path of the running install when that is a runnable file (an npm install, a compiled binary), otherwise the bare `rove`/`kobe` name resolved on `PATH`, which is what a dev checkout run through `bun` falls back to. In that fallback your callbacks run **a different build than the daemon that launched you**, so a verb or flag the daemon has can still fail as a usage error: compare `$ROVE_BIN_PATH --version` against the daemon's own `roveVersion` (see [Which host am I talking to](#which-host-am-i-talking-to)) before blaming your own arguments |
 | `ROVE_SOCKET_PATH` | daemon unix socket, for raw JSON requests |
 | `ROVE_HOME_DIR` | set when Rove runs against a non-default home (keep passing it through) |
 | `ROVE_PLUGIN_ID`, `ROVE_PLUGIN_ROOT` | who you are, where your files are |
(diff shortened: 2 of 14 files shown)


Sma1lboy:
### Harness screenshots — the frozen board, read off the terminal

Identical key sequence through `/harness` → xterm.js → PTY sidecar → real OpenTUI (fixture port base 7727, tab state and PTY host reset between the two runs so each frame has exactly one board). The fixture daemon was `kill -9`'d about 10 seconds before each capture, while the pane's hosted PTY kept running — which is the whole point, that survival is by design.

GitHub image upload isn't available from this session, so the PNGs are on disk:

| File | The two lines that matter |
|---|---|
| `/Users/jacksonc/rove-evidence/loop-r20-bi/BEFORE-daemon-killed.png` | `task:  (none)`<br>`host:  connected` |
| `/Users/jacksonc/rove-evidence/loop-r20-bi/AFTER-daemon-killed.png` | `task:  01M1PXKAHH2WNA2KC9NCH7FDZG`<br>`host:  UNREACHABLE — reconnecting` |
| `/Users/jacksonc/rove-evidence/loop-r20-bi/after-1-board-live.png` | the same board before the kill: `host:  connected`, rendering the live task list |

BEFORE is the merge-base source tree for all three packages (`packages/kobe/src`, `packages/kobe-daemon/src`, `packages/kobe-plugin-sdk/src`), rebuilt — not a stash, not a mock.

The probe plugin that draws these frames is at `/Users/jacksonc/rove-evidence/loop-r20-bi/board/`; it latches on connection loss instead of reconnecting so the captured frame is stable. The recipe documented in PLUGIN-SDK.md reconnects.

PR #896, opened 2026-09-04
A cap that calls itself generous deletes the overflow, and a 30s poll that is really 169s

Sma1lboy:
Two daemon costs that each broke a promise their own comment made. Both were
measured before and after at three or four fleet sizes, through the real code
path — a real `pty-host` process for one, real `spawn`ed children counted by a
PATH shim for the other. No screenshots: neither item has a UI surface, and
nothing rendered changes.

Re-runnable probes live in `.scratch/loop-r20-bj/`; raw output is in
`/tmp/loop-r20-bj-evidence/`.

---

## 1. The freeze store deletes scrollback it merely decided not to restore

`pty-freeze-store.ts:187` capped restores at 64 records and `rmSync`'d the
overflow. The comment justifying 64 made two claims, both false:

- *"several times any realistic number of open terminal tabs"* — the machine
  this was written on held **108 records / 69MB** (110 / 70M by the end of the
  session; it is still growing). Everything past position 64 was dated
  **Sep 3 22:43** — the previous day's work. It grows with tab churn over a
  host's lifetime, not with task count; one task there holds `tab-60`,
  `tab-63`, `tab-68`.
- *"bounds the boot read at ~32MB"* — the cap ran **after** `readFileSync` on
  every record, so the read was unbounded in file count.

**Change.** The cap is now `FREEZE_RESTORE_MAX_BYTES` (64MB), applied *before*
any read by sorting on `statSync` mtime, and a record past the budget is left
on disk rather than deleted. Only the 14-day TTL deletes, and it now reaches
records the budget never reads (mtime can only run at or ahead of a record's
own stamp, so an mtime-expired file is expired for certain) — so the directory
is still bounded. The boot reports what it restored, deferred and expired.

Bytes rather than records because bytes are what both costs actually track: the
boot read, and the rings the host then holds in memory all session.

### PASS criterion
With a directory larger than the budget, the number of records **deleted** is 0,
and the load stops scaling with directory size.

### Proof — seeded directories at the operator's mean record size (640KB)
`bun .scratch/loop-r20-bj/freeze-probe.ts`

| records | on disk | load ms before → after | restored before → after | **deleted before → after** |
|---|---|---|---|---|
| 64  | 42 MB  | 10 → 10 | 64 → 64  | 0 → 0 |
| 108 | 71 MB  | 17 → 13 | 64 → 102 | **44 → 0** |
| 200 | 131 MB | 33 → 12 | 64 → 102 | **136 → 0** |
| 400 | 262 MB | 82 → 12 | 64 → 102 | **336 → 0** |

Load time went from scaling 10→82ms with directory size to flat ~12ms: that is
the ordering fix, externally visible. Restored count rose 64 → 102 because the
budget is spent on bytes rather than on a record count guessed against a
different ring size.

### Proof — a real `pty-host` process, seeded with 120 records / 75MB
`.scratch/loop-r20-bj/live-boot.sh` (isolated home, port base 8123, run from
source; BEFORE captured by `git checkout <base> -- <the two files>`)

```
== before ==                                    == after ==
seeded : 120 records,  75M                      seeded : 120 records,  75M
(no freeze line at all)                         [pty-host freeze] restored 102 (64MB), 18 left unread past the 64MB budget
after  : 64 records,  40M                       after  : 120 records,  75M
```

**56 records destroyed by one boot, silently → 0 destroyed, and it says what it
did.** Full output: `/tmp/loop-r20-bj-evidence/live-boot-{before,after}.txt`.

---

## 2. The PR poller's documented 30s is not the rate you get

`pr-status-collector.ts:319` awaited each `gh pr list` in turn, and
`ticker.ts` drops any tick landing while a pass runs — so the real per-task
refresh was `max(30s, N × gh_latency)` while `DEFAULT_PR_STATUS_POLL_MS` and the
module doc both said 30s. This is the loop's recurring defect in a performance
costume: a documented interval that is not the real one, with nothing telling
you which you are getting.

**Change.** A fixed pool of `PR_POLL_CONCURRENCY` (8). Task selection is now a
synchronous pre-pass against one `opts.now`, so which tasks are due no longer
depends on how long the pass takes. Results land in per-task slots, so the
returned changed-ids stay in task order however the polls interleave. The tick,
the jitter, and the per-task backoffs are untouched.

### PASS criterion
Pass duration at N=200 under 30s (within one tick), down from 160s+.

### Proof — real `spawn`ed children through a counting PATH shim
`bun .scratch/loop-r20-bj/pr-pass-probe.ts` (real `makeGhPrViewRunner`, shim
`gh` at 800ms; concurrency computed from wall-clock epoch-ms start/end pairs the
shim writes, since `hrtime` is not comparable across processes)

| tasks | pass before → after | peak concurrency | real `gh` children | effective refresh before → after |
|---|---|---|---|---|
| 18  | 15.2s → **2.6s**  | 1 → 8 | 18 → 18   | 30s → 30s |
| 50  | 42.1s → **6.0s**  | 1 → 8 | 50 → 50   | 42s → **30s** |
| 200 | 169.5s → **21.5s** | 1 → 8 | 200 → 200 | 169s → **30s** |

Same number of children spawned; only the overlap changed. At 200 tasks the
pass now fits inside one tick, so the documented interval is the real one at
every size measured.

---

## Tests

`test/daemon/pty-freeze-store.test.ts` replaces the old cap test with four:
over-budget records are deferred not deleted; a lone record larger than the
whole budget still restores; a record the budget never reads still expires (the
pin on the pre-read TTL check — a budget that merely stopped reading would
strand it and let the directory grow unbounded again); and the summary reports
restored/deferred/expired separately.

Mutation-verified at two different sites, re-run after the rebase:

| mutation | result |
|---|---|
| over-budget records pushed to `stale` again (the old harm) | 2 failed / 13 passed |
| pre-read mtime TTL check deleted | 1 failed / 14 passed |

## Gates (post-rebase onto `3ef2066`)

- root `bun run lint` — clean; `bun run typecheck` — 4/4 clean
- `bun run test:fast` — 458 files, **4527 passed**
- `KOBE_INCLUDE_SOCKET=1 bun run test:socket` — 100 files, **834 passed**
- `bun test test/render` — 108 files, **479 passed**

## Not proved

- The 64MB budget is an argued number, not a measured optimum. It is bounded
  below by "restore more than the 64-record cap did on a real directory" and
  above by host RSS; I did not measure where restore latency or host memory
  actually starts to hurt, so a later round may want to move it.
- Concurrency 8 was verified to fit 200 tasks in one tick at 800ms latency. I
  did not test it against a real rate-limited `github.com` — the shim never
  leaves the machine, deliberately.

## Boundaries respected

Left alone as instructed: the scrollback-write-volume and `git status` walk
briefs, `core/base-ref-cache.ts`, `plugins/runtime.ts`, the issue store,
`refs/`, `.github/workflows/*`. The operator's `~/.rove` was read only — its
`pty-sessions` directory still holds every record it did before. My isolated
daemon and pty-host both idle-exited; verified no process holds a socket under
`.scratch/loop-r20-bj`.

### DIFF
--- a/.changeset/freeze-restore-budget.md
+++ b/.changeset/freeze-restore-budget.md
@@ -1,0 +1,27 @@
+---
+"@sma1lboy/rove": patch
+---
+
+A pty-host boot no longer deletes the scrollback it decided not to restore
+
+`loadFrozenSessions` kept the 64 newest frozen sessions and `rmSync`'d the rest.
+The cap's comment called 64 "several times any realistic number of open terminal
+tabs"; a machine measured while writing this held 108 records / 69MB, and every
+record past 64 was from the previous day's work — so the next host boot would
+have permanently deleted 44 tabs' scrollback. The count grows with tab churn
+over a host's lifetime, not with how many tasks you have: one long-lived task
+cycling `tab-60`, `tab-63`, `tab-68` gets there on its own.
+
+The cap is now a restore budget in bytes (`FREEZE_RESTORE_MAX_BYTES`, 64MB),
+and a record past it is left on disk rather than deleted. It is also applied
+*before* reading, off each file's mtime, which is what the old comment claimed
+("bounds the boot read at ~32MB") and did not do — the 64-record cap ran after
+every file had already been read, so a 400-record directory read 262MB to keep
+32MB of it. Measured on seeded directories at the same mean record size: 108
+records deleted 44 and now deletes 0; 200 deleted 136 and now deletes 0; 400
+deleted 336 and now deletes 0. Load time was 10/17/33/82ms across those four
+sizes and is now flat at ~13ms.
+
+Only the 14-day TTL deletes now, and it reaches records the budget never reads,
+so the directory is still bounded. A boot that defers or expires anything says
+so in `pty.log` — a store that quietly loses scrollback was the actual harm.
--- a/.changeset/pr-poller-concurrency.md
+++ b/.changeset/pr-poller-concurrency.md
@@ -1,0 +1,20 @@
+---
+"@sma1lboy/rove": patch
+---
+
+The PR-status poller's 30s interval is now the rate you actually get
+
+One pass awaited each `gh pr list` in turn, and the ticker drops any tick
+arriving while a pass is still running — so the real per-task refresh was
+`max(30s, N × gh_latency)` while `DEFAULT_PR_STATUS_POLL_MS` and the module doc
+both said 30s. A pass now runs up to `PR_POLL_CONCURRENCY` (8) calls at once.
+Measured through the real `spawn` path with a counting shim at 800ms per call:
+18 tasks 15.2s → 2.6s, 50 tasks 42.1s → 5.9s, 200 tasks 169.5s → 21.5s, with
+peak concurrency going from 1 to 8 and the same number of `gh` children spawned.
+At 50 tasks the effective refresh was 42s and is now the documented 30s.
+
+This matters past a stale chip: `prStatus` is the only CI truth Rove holds, and
+an unattended worker asks it whether its own PR is green. The tick, the jitter,
+and the per-task backoffs are unchanged — only the pass was serial, and `gh`
+waits on the network rather than the CPU, so serialising it was never buying
+back the subprocess budget its comment claimed.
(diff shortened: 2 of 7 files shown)


Sma1lboy:
Also updated `docs/SESSIONS.md` — it documented the old behaviour in two places ("the 64 newest records under 14 days old", and "only the **64 most recently updated** survive — past that, the oldest go"). Both now describe the 64MB restore budget, that deferred records stay on disk, and that the TTL is what deletes. `packages/kobe-docs/content/**` is generated from `docs/` and gitignored, so nothing to mirror by hand.

PR #900, opened 2026-09-04
The error path you selected is not the path you paste, and "Copied" was a guess (A, B, C, E)

Sma1lboy:
Batches **A**, **B**, **C** and **E** from `.scratch/loop-r21-text/text.md`. **Batch D is not implemented** — it is marked NEEDS OWNER SIGN-OFF and the choice between copy-on-mouse-up, a new chord, and `selectable={false}` on panes where selection is meaningless is a placement call. Its stated prerequisite is now in place: `copyTextToSystemClipboard` returns whether anything was copied, so whichever way D goes it will not ship a second surface claiming a copy it never verified.

Screenshots and command output: `/tmp/loop-r21-bm-quoll/`. Private harness port base **7623**, fixture home `.scratch/opentui-visual-7623/`; the fixture daemon **and** its PTY host were stopped and verified gone (no process holds a path under `opentui-visual-7623`).

---

## Batch A — soft wrap, one root cause, two defects

`pty-xterm-snapshot.ts` built one `TerminalRow` per xterm **grid** row and never read `line.isWrapped`. The flags now ride the snapshot (`DataListener`'s 4th argument + `captureWrapped()`, both optional so backends without an emulator are untouched), and `terminal-wrap.ts` holds the single definition of "this row continues the one above" that both consumers read.

**PASS (A1):** a wrapped line pastes as ONE line, and `copied.includes(<the path>)` is true.
**PASS (A2):** a needle straddling the wrap reads `1/1` and paints across both rows; the one-row control stays `1/1`.

### Proof — the terminal buffer, through the real `XtermSnapshotEngine`

`/tmp/loop-r21-bm-quoll/A-buffer-proof.txt` (`bun .scratch/repro-wrap-fixed.ts` from `packages/kobe`). BEFORE = calling the consumers without the flags, which is byte-for-byte the old path since both parameters are optional.

```
row 0 wrapped=false :: "ERROR in /Users/jacksonc/i/kobe/packages"
row 1 wrapped=true  :: "/kobe/src/tui/panes/terminal/terminal-se"
row 2 wrapped=true  :: "arch.ts:41:9"

A2  needle "ckages/kobe/" (present on screen: true)
    BEFORE findMatches(rows, needle)          -> 0 match(es)
    AFTER  findMatches(rows, needle, wrapped) -> 1 match(es), spans rows 0..1
    control (needle inside one row)           -> 1 match(es)

A1  BEFORE copied: "ERROR in /Users/jacksonc/i/kobe/packages\n/kobe/src/tui/panes/terminal/terminal-se\narch.ts:41:9"
    AFTER  copied: "ERROR in /Users/jacksonc/i/kobe/packages/kobe/src/tui/panes/terminal/terminal-search.ts:41:9"
    BEFORE newline the user did not select: true    pasteable path survives: false
    AFTER  newline the user did not select: false   pasteable path survives: true
```

### Proof — `/harness` → xterm.js → PTY sidecar → real OpenTUI

| | BEFORE | AFTER |
|---|---|---|
| straddling needle `ROVEWRAP` | `A2-before-straddle.png` — `/ROVEWRAP  no matches`, with `…XXXROVE` ending one row and `WRAPZZZ…` starting the next | `A2-after-straddle.png` — `/ROVEWRAP  1/1`, hit painted in accent across BOTH rows |
| control needle `WRAPZ` (fits one row) | `A2-before-control.png` — `1/1` | `A2-after-control.png` — `1/1`, unchanged |
| drag highlight over the wrapped line | `A1-before-highlight.png` | `A1-after-highlight.png` — identical two-row highlight |

The highlight pair is the "**do not change the highlight**" check: you selected two visual rows and still see two highlighted rows; only the extraction joins. The drag driver (`packages/kobe-harness/.scratch/drag-highlight-shot.ts`, gitignored) deliberately screenshots mid-drag and **never releases** — the release is what fires copy-on-select, which on this machine pipes into the operator's real `pbcopy`.

### Regression tests (mutation-verified)

`test/tui/terminal-wrap.test.ts` (10 tests). Mutation 1 — `isWrapContinuation` → always false: 5 red. Mutation 2, different site — `rowAtOffset` always answers the first row: 2 red. Both restored, all green.

`overlayMatches` now compares a match's whole span against the viewport instead of its anchor alone: a hit across a wrap starts one row above the visible window and still has a tail inside it.

---

## Batch B — the answer is no longer discarded

`copyTextToSystemClipboard` returns `Promise<boolean>`: the local pipe's exit status is read, and the OSC 52 writer is typed `(text) => boolean | undefined` so opentui's return value can survive the signature. `copyTaskFieldFlow` awaits it and toasts the failure instead of `Copied branch X`. `copySelection` stays silent on success (a toast per drag would be noise) and speaks on failure.

**PASS:** a copy where both channels refuse says so.

### Proof — a genuinely failing channel, measured

`/tmp/loop-r21-bm-quoll/B-clipboard-proof.txt`. Nothing here writes to the operator's clipboard — `cat > file` stands in for `pbcopy` on the identical spawn path:

```
resolveClipboardCopyCommand:
  darwin (all binaries present)              -> ["pbcopy"]
  darwin (no binaries present)               -> null
  linux  (headless: no wl-copy/xclip/xsel)   -> null
  win32                                      -> ["clip"]      (was null)

sh -c <missing binary> exit code: 127        (Bun.spawn does NOT throw)
pipeToClipboardCommand(text, ["sh","-c","rove-no-such-clipboard-binary"]) -> false
pipeToClipboardCommand(text, ["sh","-c","exit 3"])  (xclip w/o $DISPLAY)  -> false
pipeToClipboardCommand(text, null)                  (headless Linux)      -> false
pipeToClipboardCommand(text, ["sh","-c","cat > f"]) (a channel that works)-> true, file holds "feat/some-branch"
```

`test/tui/clipboard-copy.test.ts` pins all four against real processes.

### Proof — the toast, through the harness

Both shots were taken with the SAME two forced-failure lines applied — `darwin` resolving to `rove-no-such-clipboard-binary`, and the OSC 52 writer returning `false` (what `copyToClipboardOSC52` returns on Terminal.app). That reproduces the described environment on a mac; the only difference between the two builds is the fix. Neither line is in this PR (`grep -rn TEMP-EXPERIMENT packages/kobe/src` → nothing).

| BEFORE | AFTER |
|---|---|
| `B-before-copy-lie.png` — green **✓ Copied branch visual-fixture** over a clipboard neither channel reached | `B-after-copy-failed.png` — red **✕ Couldn't reach a clipboard — no clipboard command on PATH, and this terminal refused OSC 52.** |

Worth recording: `B-probe-osc52.png` is the same build with ONLY the local pipe forced to fail, and it correctly still says **Copied** — xterm.js accepts OSC 52, so one channel really did take the text. The fixed code is not just pessimistic.

### B2 — Windows

`resolveClipboardCopyCommand` now returns argv rather than a shell string, because there is no `sh` on Windows to hand a command line to; the spawn goes direct. `win32` resolves `clip`, falling back to `powershell -NoProfile -Command Set-Clipboard`. The `where` probe already written for Windows is now reachable. `darwin` is probed like every other platform — a mac without `pbcopy` now reports "no clipboard" rather than spawning a command that fails where nobody reads the status.

### B3 — whitespace-only selection

Kept as a deliberate skip, now stated in the code: a stray click is a zero-width selection, and clobbering the clipboard with a run of spaces on every pointer twitch is worse than doing nothing. Nothing is reported because nothing was attempted.

---

## Batch C — the parked hit is addressed by line id

`index` was an array position kept across recomputes. `ParkedHit` now records the absolute line id (+ column) the viewport is anchored by, and `index` is re-derived from the recomputed list each frame. A trimmed line falls forward to the next surviving hit; an epoch change (a resize reflows history) **drops** the park rather than mis-mapping it — the same discipline `followWindowShift` applies to a selection. Backends with no stable line ids keep the positional form, explicitly named as such.

**PASS:** after a trim the reported index still names the same occurrence.

**Proof — mutation-verified unit test**, `test/tui/terminal-search-park.test.ts` (7 tests). Mutation: `resolveParkedIndex` returns the array position (the pre-change behaviour) → 4 red, including "keeps naming the SAME occurrence after a scrollback trim". Second mutation at a different site (identity match removed + epoch guard dropped) → 2 red.

**Could NOT prove with screenshots.** The drift needs a saturated scrollback plus output arriving while the query row is open, and its symptom is "the counter says 3/5 while the accent sits on a different occurrence" — which a still frame cannot distinguish from a correct 3/5. I did not stage it in the harness and am not claiming a screenshot pair for C.

---

## Batch E — one stale comment

`use-terminal-search.ts` documented `shift+enter` walking back through history. The bindings are `terminal.search.older: ["up", "return"]` / `terminal.search.newer: ["down"]`, and `keybindings-table.ts:395` explains why the chord cannot exist (without the kitty protocol a terminal sends the same CR for both). Comment corrected; keys untouched. No chord added or moved anywhere in this PR.

---

## One thing I got wrong

The first run of my Batch B evidence script called `copyTextToSystemClipboard` directly, which resolves the REAL platform command — so it ran `pbcopy` once and left the string `x` on this machine's clipboard. That crossed the "do not write to the operator's clipboard" line. The script no longer calls it (the composition is covered by tests instead), and every other clipboard claim here is measured against `cat > file` or the resolver's return value.

## Gates

`bun run lint` · `bun run typecheck` · `bun run test:fast` (461 files / 4551 tests) · `bun test test/render` (479) · `KOBE_INCLUDE_SOCKET=1 bun run test:socket` (100 files / 834) — all green, rebased on `origin/main` @ `cf7fb5b75`.

### DIFF
--- a/.changeset/wrapped-lines-and-honest-clipboard.md
+++ b/.changeset/wrapped-lines-and-honest-clipboard.md
@@ -1,0 +1,42 @@
+---
+"@sma1lboy/rove": patch
+---
+
+Copying a soft-wrapped terminal line pastes as one line, and search can find text that straddles the wrap
+
+The terminal snapshot builds one row per GRID row and never read xterm's
+`isWrapped`, so a line the emulator broke across columns arrived downstream as
+several unrelated rows. Two things a user does constantly fell out of that.
+Drag-selecting a path out of a build log copied
+`…/packages\n/kobe/src/…\narch.ts:41:9` — three broken shell commands where one
+path was selected. And the `/` scrollback search, which ran `indexOf` per row,
+reported `no matches` for a needle spanning the break: a confident negative for
+text visible two rows above the query. xterm.app, iTerm2 and tmux all rejoin
+soft-wrapped rows; Rove was the outlier.
+
+The flags now travel with the snapshot, and both consumers group rows into
+logical lines from them. The highlight is unchanged — you selected two visual
+rows and still see two highlighted rows; only the extraction joins. A search hit
+across a wrap is a multi-row range, which the existing paint already handles.
+
+**Clipboard writes now report whether anything was copied.**
+`copyTextToSystemClipboard` returned `void` and discarded both channels'
+answers: the local pipe's exit status was never read (a missing command exits
+127 without throwing; `xclip` with no `$DISPLAY` passes the `which` probe and
+then fails), and OSC 52's boolean could not survive a `(text) => void`
+signature. So on a headless box with no `wl-copy`/`xclip`/`xsel`, in a terminal
+that refuses OSC 52, **Copy branch name** toasted `Copied branch feat/whatever`
+over an untouched clipboard. It now says it could not reach a clipboard, and the
+pane's copy-on-select — silent on success, as it should be — speaks on failure.
+Windows also gets a local clipboard command for the first time (`clip`, then
+PowerShell's `Set-Clipboard`): the `where` probe written for it was unreachable
+behind a platform check that returned null, leaving OSC 52 as its only channel.
+
+**The parked search hit no longer drifts out from under its counter.** The hit
+you walked to was remembered as a position in the match array, which a
+scrollback trim renumbers — the counter kept reading `3/5` while the accent
+highlight had moved to a different occurrence, and `enter` walked on from the
+wrong place. It is now remembered by absolute line id, re-derived each frame,
+falling forward to the next surviving hit when that line is trimmed and dropped
+outright when a resize resets line numbering — the same discipline
+`followWindowShift` already applies to a selection.
--- a/packages/kobe/src/lib/clipboard-command.ts
+++ b/packages/kobe/src/lib/clipboard-command.ts
@@ -4,21 +4,29 @@ import { spawnSync } from "node:child_process"
 
 export type ClipboardProbe = (binary: string) => boolean
 
-const LINUX_CLIPBOARD_CANDIDATES: readonly { readonly binary: string; readonly command: string }[] = [
-  { binary: "wl-copy", command: "wl-copy" },
-  { binary: "xclip", command: "xclip -selection clipboard -in" },
-  { binary: "xsel", command: "xsel --clipboard --input" },
-]
+/** Argv, not a shell string: Windows has no `sh` to hand a command line to. */
+const CLIPBOARD_CANDIDATES: Readonly<Record<string, readonly (readonly string[])[]>> = {
+  darwin: [["pbcopy"]],
+  linux: [["wl-copy"], ["xclip", "-selection", "clipboard", "-in"], ["xsel", "--clipboard", "--input"]],
+  // Windows is a supported platform, and without this its only clipboard
+  // channel is OSC 52 — which the terminals shipping on Windows may refuse.
+  // `clip` has been in every Windows since XP; PowerShell's `Set-Clipboard`
+  // is the fallback for a stripped image.
+  win32: [["clip"], ["powershell", "-NoProfile", "-Command", "Set-Clipboard"]],
+}
 
+/**
+ * The argv that pipes stdin into this platform's clipboard, or null when the
+ * platform has none on PATH. `pbcopy` is probed like everything else: a macOS
+ * without it should say "no clipboard" rather than spawn a command that fails
+ * where nobody reads the exit code.
+ */
 export function resolveClipboardCopyCommand(
   platform: NodeJS.Platform | string,
   hasCommand: ClipboardProbe,
-): string | null {
-  if (platform === "darwin") return "pbcopy"
-  if (platform === "linux") {
-    for (const candidate of LINUX_CLIPBOARD_CANDIDATES) {
-      if (hasCommand(candidate.binary)) return candidate.command
-    }
+): readonly string[] | null {
+  for (const candidate of CLIPBOARD_CANDIDATES[platform] ?? []) {
+    if (hasCommand(candidate[0] as string)) return candidate
   }
   return null
 }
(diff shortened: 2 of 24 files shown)


Sma1lboy:
Screenshots (absolute paths on this machine — `/tmp/loop-r21-bm-quoll/`):

**Batch A2 — search across a soft wrap**
- BEFORE `/tmp/loop-r21-bm-quoll/A2-before-straddle.png` — `/ROVEWRAP  no matches`
- AFTER  `/tmp/loop-r21-bm-quoll/A2-after-straddle.png` — `/ROVEWRAP  1/1`, painted across both rows
- control BEFORE `/tmp/loop-r21-bm-quoll/A2-before-control.png` · AFTER `/tmp/loop-r21-bm-quoll/A2-after-control.png` — both `/WRAPZ  1/1`

**Batch A1 — the highlight must not change**
- BEFORE `/tmp/loop-r21-bm-quoll/A1-before-highlight.png` · AFTER `/tmp/loop-r21-bm-quoll/A1-after-highlight.png`

**Batch B — the toast**
- BEFORE `/tmp/loop-r21-bm-quoll/B-before-copy-lie.png` — green `✓ Copied branch visual-fixture`
- AFTER  `/tmp/loop-r21-bm-quoll/B-after-copy-failed.png` — red `✕ Couldn't reach a clipboard …`
- probe  `/tmp/loop-r21-bm-quoll/B-probe-osc52.png` — local pipe forced to fail, OSC 52 accepted, still says Copied

**Command output**
- `/tmp/loop-r21-bm-quoll/A-buffer-proof.txt` · `/tmp/loop-r21-bm-quoll/B-clipboard-proof.txt`

PR #943, opened 2026-09-06
Six things missing when you watch five sessions at once

Sma1lboy:
Seven things a person watching five sessions at once did not have. Six landed;
the seventh turned out to already exist. One commit each.

Each item below carries a BEFORE/AFTER pair taken through the real
`/harness` → xterm.js → PTY → OpenTUI path, on an isolated fixture
(`KOBE_VISUAL_PORT_BASE=5673`). BEFORE shots were taken with
`git checkout origin/main -- packages/kobe/src packages/kobe-daemon/src` and
the fixture daemon + PTY host killed by pidfile, so they are the real old
code, not a mock.

---

### 1. "How many are waiting on me" — ALREADY SHIPPED, no change made

The brief called for a count in the permanent chrome, on the grounds that
`attentionCount` had four hits and none in the sidebar. That grep misses the
one that matters: `host-sidebar-mount.tsx` already renders
`` `${t("workspace.inbox.title")} ${inbox.counts.total}` `` in the sidebar's
FIRST row, bold + `warningOnHost` whenever it is above zero, and
`inbox.counts.total` is exactly the pending-episode count.

Every screenshot in this PR shows it: `ROVE INBOX 2` in the top-left corner.

I did not add a second one. A per-project `!N` on the section headers would be
a second control for the same fact — the thing `nav-core.ts`'s own comment
says the rail deliberately avoids.

- Evidence: `.scratch/shots-butterfly/after-states.png` (top-left, `INBOX 1`),
  `after-inbox.png` (`INBOX 2` beside the open pane).

### 2. F7 mixed "done" and "stuck" into one queue

`sortAttentionInbox` was pure age order, so a permission prompt queued behind
every turn that had merely finished — and F7 clears each item as it visits it,
so reaching the stuck one dismissed the others on the way.

Blocked episodes now sort ahead of plain completions; oldest-first is preserved
inside each band. Every state except `turn_complete` bands as blocking, so a
new episode kind defaults to the side where being wrong costs one keypress
rather than burial.

The pair uses the DISCRIMINATING seed — the blocked episode is the NEWEST, so
only the band can float it:

| | |
|---|---|
| BEFORE `before-inbox.png` | `✓ done 3s` on top, `? needs input 1s` below |
| AFTER `after-inbox.png` | `? needs input 1s` on top, `✓ done 3s` below |

Tests: `test/tui-react/attention-inbox-order.test.ts`. Two mutations at
different sites (drop the band; drop the within-band age order) each go red.
`docs/TUI.md` Inbox section updated.

### 3. Rows did not say how long

Nothing in the rail dated a row, so a tab stuck for forty minutes and one
stuck for forty seconds looked identical.

A row that is WORKING or STOPPED now carries `relativeAge(activity.at)` beside
its label. Quiet rows get nothing. No new timer — the tree already re-renders
on its ~2s branch tick, and the label is derived outside `useTabRowBaseView`'s
memo so idle rows still rebuild nothing.

| | |
|---|---|
| BEFORE `before-states.png` | `! claude 1`, `! generic 1` |
| AFTER `after-states.png` | `! claude 1  5s`, `! shell 1  5s` |

The frame goldens' fixture instant became a fixed AGE — a hard-coded epoch
would render a number that grows daily and re-break them on a schedule.

### 4. Nothing could sort by "who needs me" — **CHORD SEMANTICS CHANGE**

**This is the one that needs your call.** `t` was a two-state toggle
(default ↔ recent). It is now a three-state cycle: **default → recent →
attention → default**. No new key, no moved key; the same key does one more
thing. `docs/KEYBINDINGS.md` and `docs/CONFIGURATION.md` updated to match.
Say the word and I will split it back out.

`attention` puts stopped tasks at the top of each project, then unread
completions, then the rest — most-recently-touched inside each band.

The fixture is seeded so the blocked task is LAST in stored order, so the sort
has something to move:

| | |
|---|---|
| BEFORE `before-sort-attention.png` | after `t` `t`: unchanged — the toggle came back to `default` |
| AFTER `after-sort-attention.png` | after `t` `t`: `flaky-test` (blocked) floated from last to first |
| (`after-sort-default.png` is the same fixture before the presses) |

`compareAttention` borrows `compareRecent` for its within-band tiebreak rather
than restating it; `compareRecent` itself is untouched, since the Inbox's
RECENT section shares it and must not learn about engine activity.
`buildTreeRows` takes a state READER, not the daemon's map, so it stays pure.
The `ui-prefs` wire contract widens too, or a persisted `attention` would be
silently rewritten to `default` by the daemon's own normalizer.

`tree-core.ts` crossed the 500-line cap, so row NAMING and ADDRESSING moved to
`tree-row-naming.ts` (nothing there reads the task list). All names
re-exported; no call site moved.

Test: `test/render/sidebar-sort-key.test.tsx` drives the REAL
`useSidebarHostState` cycle through a real `t` keypress and walks all three
stops. Dropping `attention` from the cycle turns it red.

### 5. The landing cue was on a surface that is off by default

The turn-complete pulse lived only in `tab-strip.tsx`, and
`DEFAULT_TAB_STRIP_MODE` is `"never"` — so on a stock install nothing
announced a landed turn. Tab rows now flash the same 600ms, and both surfaces
read one `DONE_PULSE_MS`.

A one-shot 600ms window cannot be timed across two processes, so the shots
were taken under a loop firing a fresh completion stamp every 400ms. Sampled
pixels from the `claude 1` row:

| shot | glyph | label |
|---|---|---|
| AFTER pulsing `after-pulse-on.png` | **(110,142,96)** green | **(234,231,223)** bright, BOLD |
| AFTER resting `after-pulse-off.png` | (120,116,110) grey | (169,163,154) muted |
| BEFORE, same loop `before-pulse-on.png` | — | (169,163,154) muted |

The label channel is the apples-to-apples one: under a continuous stream of
completions the old code never brightens; the new one does.

**Caveat, stated plainly:** the BEFORE shot's glyph shows a different activity
state (the fixture's arbitration kept a sticky one), so the glyph column is
not a fair comparison there — only the label column is. The mechanism itself
is covered by `test/render/sidebar-done-pulse.test.tsx`, whose two mutations
(drop the mount seed guard; drop the stamp equality) each go red.

### 6. `rate_limited` was red on the rail and amber everywhere else

A wall that clears itself when the window rolls looked exactly like a dead
engine. `tab-strip.tsx`'s `turnColor` and the Inbox's `itemColor` have always
drawn it amber.

Sampled pixels, same glyph, same screenshot:

| | `rate_limited` `!` | `permission_needed` `!` |
|---|---|---|
| BEFORE `before-states.png` | (212,117,99) | (212,117,99) — identical |
| AFTER `after-states.png` | **(232,201,107)** amber | (212,117,99) red — unchanged |

The `!` glyph is deliberately untouched (`row-view.ts` already recorded the
decision not to split it), and `activityBadgeFor` now asks `activityToneFor`
instead of keeping a second copy of the state list.

### 7. Kanban cards could not be moved

`enter` → drawer → `tab` across fields → `esc` was the only human route out of
a column. Cards now take the sidebar's right-click gesture, through the same
`ContextMenu`, routing to the same `setStatus` op.

| | |
|---|---|
| BEFORE `before-kanban-menu.png` | the same right-click selects the card and opens nothing |
| AFTER `after-kanban-menu.png` | `status → doing / hold / done` (current status `open` correctly absent) |
| AFTER `after-kanban-moved.png` | after `enter`: Backlog (0), In progress (2) |

No new chord — the menu is mouse-opened and answers up/down/enter/escape,
which it owns itself so the page's bindings stand down while it is up.
Test: `test/render/kanban-card-menu.test.tsx`; removing the right-click branch
turns two of four red.

---

## Left for you (from the brief's "don't do" list)

- **Wide-terminal "back to the last one" recent row** (`host-pages.tsx` narrow
  gate). Costs a row and cuts against the density preference — your call.
- **Multi-select** across sidebar / Kanban / Worktrees. One selection model,
  three surfaces; not one PR.
- **No visible indicator of which sort mode is active.** Three silent states
  are harder to read than two were. The row ORDER is still the feedback (as it
  was for the two-state toggle) so I did not add a chip, but if `t` stays
  three-state a label may be worth it.

## Residual risks

- `attention` sort re-runs `buildTreeRows` on every daemon activity push while
  that mode is selected. The other two modes are unaffected (the engine map is
  excluded from their memo deps), but a very large task list in `attention`
  mode does more work per push than before.
- The three-state `t` writes `activeSortMode: "attention"` to `state.json`. An
  OLDER Rove build reading that file falls back to `default` (it is not in its
  union) rather than failing — checked, but it means downgrading loses the
  setting silently.
- Sentinel/shell tabs in the fixture carry activity glyphs, which is why the
  shots show `shell 1` rows with `!`. That is fixture shape, not a product
  change.

## Verification

`bun run test:fast` 4921 ✓ · `bun test test/render` 522 ✓ ·
`KOBE_INCLUDE_SOCKET=1 bun run test:socket` 956 ✓ · root `bun run lint` clean ·
`tsc --noEmit` clean · `check-i18n` 946 keys × 2 locales in sync ·
one `patch` changeset.

(The package-level `bun run lint` reports 3 pre-existing errors in
`test/daemon/worktree-probe.test.ts` and `test/engine/ps-probe-deadline.test.ts`
— neither file is touched here, and CI runs the repo-root lint, which is clean.)

## CI: `windows` took three runs — evidence it is the runner, not this PR

All ten checks are green now. `windows` failed twice first, and it is worth
writing down why I concluded it was not mine, because "rerun went green" alone
proves nothing:

1. **A DIFFERENT test failed each run.** Run 1:
   `test/tui/terminal-pty-lazy-snapshot.test.ts` ("keeps the eager push path for
   live subscribers"). Run 2: `test/orchestrator/land-commit-failure.test.ts`
   ("squash surfaces git's own error"). Run 3: green. A regression fails the
   SAME test every time.
2. **Neither failing test shares a single module with this diff.** Checked
   import by import: `pty-xterm-base.ts`, `pty-types.ts`, `exec-host.ts`,
   `orchestrator/errors.ts`, `orchestrator/land.ts`, `worktree/exec-deps.ts`,
   `types/task.ts` — none is in the 40 files here.
3. **Both have a named wall-clock race.** The PTY one asserts a subscriber saw
   pushed content after `settle()`, which is a bare `setTimeout(…, 60)` — a
   loaded Windows runner can miss 60ms for an async xterm write. The land one
   shells out to real `git`.
4. **`main`'s `windows` job is green on its last five completed runs**, so
   neither test is chronically broken there, and each failing run still passed
   3943 of 3944 tests.

Both files pass locally on this branch.


Sma1lboy:
Relanded on `main` (was 61 commits behind, 12 conflict paths). Squash-merged onto `origin/main` and force-pushed as one commit rather than rebased — this repo only squash-merges, so the per-commit history had nowhere to land.

Three real conflicts (the other nine paths auto-merged):

- **`tree-core.ts`** — both sides extracted the same row-naming code. #974-era `main` split it into `tree-ids.ts` + `tree-labels.ts`; this branch had split it into one `tree-row-naming.ts`. The four shared functions were behaviourally identical, so `main`'s finer split wins and `tree-row-naming.ts` is dropped. Its test survives, repointed at `tree-core` (which re-exports both) and renamed `sidebar-tree-naming.test.ts`.
- **`use-tree-state.ts`** — pure union: `main`'s `machines`, this branch's `sortEngineState`, both in the memo deps.
- **`attention-inbox-core.test.ts`** — git refused a text merge (there are literal NUL bytes in `attentionInboxKey` string literals on `main`). Resolved by hand: took `main`'s file and re-applied the block that moved to `attention-inbox-order.test.ts`.

**Semantic drift caught after the conflicts were clean**, exactly as expected: the Inbox banding enumerated `prompt_deferred` / `prompt_expired`, which #959 deleted along with the whole deferred-prompt mechanism. Rather than trim two entries, the predicate is inverted to name the single exception (`turn_complete` trails, everything else blocks) — which is what its own doc comment already promised, and which an allow-list quietly broke for any state added later. Its test now derives the list from the daemon's `ATTENTION_INBOX_STATES` instead of restating it. Same removal applied to `docs/TUI.md` and the changeset.

Gates green locally on the pushed commit: root `lint`, `typecheck`, `test:fast` (5083), `bun test test/render` (540), `test:behavior` (33), `file-size-check.sh`, `coverage-gate.mjs`.

Title corrected from "Seven" — six changes ship, matching the changeset.

Sma1lboy:
Amended and re-pushed (`ab33bf8` -> `29f0d7f`) with one more change.

The reason git called `attention-inbox-core.test.ts` a binary file is that `attentionInboxItemKey` joins on a NUL byte, and the tests asserting on those keys had that raw byte sitting inside their string literals. Moving that block into `attention-inbox-order.test.ts` moved the hazard with it -- the new file came out as `data` to `file(1)` and would have refused a text merge exactly the same way the next time anyone touched it.

Those three assertions now spell the separator as a backslash-`u0000` escape sequence, which is how `attention-contracts.ts` writes it in the first place. Same strings, same assertions (37 tests green), and a repo-wide sweep now finds no `.ts` / `.tsx` file carrying a raw NUL.

Sma1lboy:
Correction to my note above: the closing claim that a repo-wide sweep finds no `.ts` / `.tsx` file carrying a raw NUL is wrong. My sweep used a shell glob that never descended into the daemon package.

`git grep -P '\x00'` finds one more: `packages/kobe-daemon/src/daemon/work-items.ts:219` joins on a raw NUL, and that file is `data` to `file(1)` — so it carries exactly the same unmergeable-conflict hazard, and a future conflict there will present as `warning: Cannot merge binary files` with no conflict markers to resolve.

Everything else in that note stands: both inbox test files are plain text now, and the three assertions use the escape sequence. `work-items.ts` is outside this PR's scope and untouched here.

PR #956, opened 2026-09-07
fix: rove update runs on Windows

ZHallen122:
## The bug

`rove update` — and the TUI's update chip (`U` on the update page, and the versions browser) — spawned bare `sh`:

- `src/cli/update.ts` → `updatePlan()` returned `{ command: "sh", args: ["-c", "curl -fsSL … | sh"] }`
- `src/tui-react/component/run-updater.ts` → `spawnSync("sh", ["-c", command])`

Windows has no `sh` on `PATH`. A default Git for Windows install adds only `…\Git\cmd` (git.exe and friends); `sh.exe`/`bash.exe` live in `…\Git\bin`. So a Windows user got:

```
rove update: failed to run sh: spawn sh ENOENT
```

on a platform Rove otherwise supports (Windows PTY host, CIM process walk, Git Bash engine launches are all wired). The only route left was `npm install -g @sma1lboy/rove@latest` by hand, which nothing told them about.

## The fix

Not a second Windows dialect of the script — `scripts/update.sh` needs no changes at all. Under Git Bash `command -v rove` lands on npm's shell shim, the `lib/node_modules` prefix derivation finds nothing and falls back to a plain `npm install -g`, which is the correct install on Windows.

New `src/lib/updater-shell.ts` points the updater at the shell Rove already uses everywhere else:

- `updaterShell()` — bare `sh` on POSIX (deliberately **not** `$SHELL`: a login fish/zsh is not the dialect the script is written in), and on win32 delegates to `resolveLoginShell()` from `kobe-daemon/daemon/platform-shell`, the one module that knows where Git for Windows puts bash and why System32's `bash.exe` (the WSL launcher) must never be used.
- `updaterShellFailureHint()` — on win32, turns a raw ENOENT into "install Git for Windows" + the manual `npm install -g` command.

Both call sites use it, so the chip and the CLI can't drift apart.

## Not changed (surfacing, not fixing)

`src/tui/lib/editor-launch.ts`, `src/tui-react/workspace/use-file-open-actions.ts` and `src/cli/config-cmd.ts` also hardcode `sh` for editor/diff/config launches. Same class of Windows bug, different slice — say the word and I'll take it in a follow-up.

`rove doctor`'s win32 block checks for Node but not for Git Bash; adding that check is also a reasonable follow-up.

## Test

- `test/lib/updater-shell.test.ts` (new) — POSIX stays `sh` even with `$SHELL=fish`; win32 resolves `C:\Program Files\Git\bin\bash.exe` with an injected `exists`; missing Git for Windows still names the real path rather than bare `bash.exe`; the hint is win32-only.
- `test/cli/update.test.ts` — expectations now go through `updaterShell()` so the Windows CI job passes too.
- `bun run test:fast`: 500 files / 4946 passed. typecheck + biome clean.

### DIFF
--- a/.changeset/windows-self-update.md
+++ b/.changeset/windows-self-update.md
@@ -1,0 +1,5 @@
+---
+"@sma1lboy/rove": patch
+---
+
+`rove update` works on Windows. The updater spawned bare `sh`, which a default Git for Windows install does not put on `PATH` — only `…\Git\cmd` is added, while `sh.exe` lives in `…\Git\bin` — so both the CLI and the TUI's update chip died with a raw `spawn sh ENOENT`, leaving `npm install -g` by hand as the only route. The updater now runs the install script through the same Git Bash every engine and terminal tab already launches through, and a missing one is reported as "install Git for Windows" with the manual command, instead of an ENOENT naming a shell.
--- a/docs/CLI.md
+++ b/docs/CLI.md
@@ -58,6 +58,11 @@ rove updates using whichever package manager owns the `rove` on your `PATH`,
 so the new version can't land in a shadowed prefix. Manual fallback:
 `npm install -g @sma1lboy/rove@latest` (or `@nightly`).
 
+The update script is POSIX shell. On Windows, Rove runs it through Git for
+Windows' bash — the same shell every engine and terminal tab launches through
+— so `rove update` needs Git for Windows installed, exactly like the rest of
+the app. Without it the command says so and points at the manual fallback.
+
 Some versions are marked breaking. Installing across one prints a heads-up
 (`dry-run` included, so the rehearsal shows it too), and the next launch asks
 you to run `rove reset` first. Worktrees are never touched.
--- a/docs/TROUBLESHOOTING.md
+++ b/docs/TROUBLESHOOTING.md
@@ -56,8 +56,9 @@ Windows needs three separate runtimes:
 - **Node.js** runs the Windows Hosted PTY process. A Bun-only global install
   does not install Node for you.
 - **Git for Windows**, including its Git Bash, supplies the POSIX shell used by
-  every engine and terminal launch. Rove deliberately does not use the WSL
-  `bash.exe`, because it cannot address the Windows worktree correctly.
+  every engine and terminal launch — and by `rove update`, whose install
+  script is POSIX shell. Rove deliberately does not use the WSL `bash.exe`,
+  because it cannot address the Windows worktree correctly.
 
 Start with:
 
--- a/packages/kobe/src/cli/update.ts
+++ b/packages/kobe/src/cli/update.ts
@@ -16,6 +16,7 @@
  */
 
 import { spawnSync } from "node:child_process"
+import { updaterShell, updaterShellFailureHint } from "../lib/updater-shell.ts"
 import {
   BREAKING_VERSIONS,
   CURRENT_VERSION,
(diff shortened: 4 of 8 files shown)


ZHallen122:
## Verified on a real Windows install (not just injected-platform unit tests)

Tested from WSL against the host Windows side, which has Git for Windows, Node, bun, and rove 0.9.172 installed (`~/.bun/bin/rove.exe`).

**The bug, on the shipped build:**

```
> rove update
rove 0.9.172 -> latest
running: curl -fsSL https://raw.githubusercontent.com/Sma1lboy/rove/main/scripts/update.sh | sh
rove update: failed to run sh: Executable not found in $PATH: "sh"
```

`where.exe sh` finds nothing. Note `where.exe bash` DOES resolve — to System32's WSL launcher, which is exactly the trap `platform-shell.js` documents: had the fix reached for bare `bash`, the install would have landed in a Linux prefix the Windows PATH never sees.

**The fix, patched sources running under Windows bun:**

```
platform       : win32
updaterShell() : C:\Program Files\Git\bin\bash.exe
updatePlan()   : {"command":"C:\\Program Files\\Git\\bin\\bash.exe","args":["-c","curl -fsSL …/update.sh | sh"], …}
spawn status   : 0 (no error)
```

Inside that shell, everything `update.sh` needs is present: `npm` → `/c/Program Files/nodejs/npm`, `curl` → `/mingw64/bin/curl`, `mktemp` → `/usr/bin/mktemp`.

**End-to-end through the real script on Windows** (`--list` mode, installs nothing) returned the published version list, exit 0.

**The script's own Windows logic**, exercised with stubbed package managers so nothing installed:

```
command -v rove : /c/Users/…/.bun/bin/rove
Updating @sma1lboy/rove: rove 0.9.172 -> v9.9.9 (via bun)
STUB npm view @sma1lboy/rove version
STUB bun install -g --no-cache @sma1lboy/rove@latest
```

It correctly detects the bun-owned install and would run the right command — confirming `update.sh` itself needs no changes. (The stub claimed latest was 9.9.9, so the verify step then flagged a mismatch, i.e. that check works too.)

**Ubuntu, same code:** `updaterShell()` → `sh`, hint → `null`, spawn status 0, and `update dry-run` / `update 0.9.171 dry-run` print the unchanged commands. POSIX behaviour is untouched.

Not run: a real full install on Windows, which would actually reinstall the package on that machine. Everything up to and including the exact `bun install -g` invocation is verified.

ZHallen122:
## Real full install on Windows — completed, exit 0

Ran the patched `runUpdateSubcommand([])` (the whole code path, not a hand-rolled spawn) under Windows bun. No stubs, real network, real package manager:

```
running: curl -fsSL https://raw.githubusercontent.com/Sma1lboy/rove/main/scripts/update.sh | sh
██████╗  ██████╗ ██╗   ██╗███████╗
…
Updating @sma1lboy/rove: rove 0.9.172 -> v0.9.172 (via bun)
✓ rove 0.9.172 -> rove 0.9.172
Thanks for using Rove. Happy building.

installed. Two background processes are still running the old build:
  daemon    → `daemon restart` (safe; never touches live sessions)
  pty host  → only `reset` replaces it …
EXITCODE=0
```

Every stage ran on Windows: Git Bash spawn → curl → the hosted script → duplicate-install scan → `bun install -g` → `hash -r` → the `rove -v` verify (`✓ rove 0.9.172 -> rove 0.9.172`) → the post-install follow-up note → exit 0.

Before/after on the machine are identical and healthy — `rove -v` → `rove 0.9.172`, `where rove` → `C:\Users\zhall\.bun\bin\rove.exe`, and `rove update dry-run` still works from the reinstalled binary. (Same version in and out because 0.9.172 is current latest; the install itself really ran.)

One harness artifact worth naming so the log above is not misread: my driver printed `kobe …` rather than `rove …` because `activeCliName()` reads the env var the real bin wrapper sets, and `bun drive-install.ts` is not that wrapper. The shipped binary prints `rove` — visible in the `rove update dry-run` line above.

PR #957, opened 2026-09-07
fix: editor launches work on Windows (rove config, file tree enter)

ZHallen122:
Stacked on #956 — it reuses that PR's shell helper, so **merge #956 first**. GitHub retargets this to `main` when #956 lands.

Closes the cluster in issue store #2.

## The bug

`binaryAvailable()` (`src/tui/lib/editor-launch.ts`) probes with `Bun.spawn(["sh", "-c", "command -v <bin>"])` and catches the throw into `false`. Windows has no `sh` on PATH, so **it reported every editor as missing** — verified on a real Windows install:

```
> set EDITOR=notepad && rove config
rove config: no editor found — set $EDITOR (or Settings → General → Editor), or edit directly:
  C:\Users\zhall\.config\rove\state.json
exit=1
```

`$EDITOR` is set, `notepad` exists. The one action the message names cannot fix it, because the gate is the probe.

Two consequences, different shapes:

- **`rove config` is unusable on Windows.** No path through it works; hand-editing `state.json` is the only route.
- **The in-TUI editor tab never appears.** `resolveEditorLaunch()` gates on the same probe, so it always returns `null` regardless of `$EDITOR` or Settings → Editor. FileTree `enter` falls through to `openExternally()`, which IS Windows-correct — so files do open, in the OS default app. Rove just looks like it has no editor tab on Windows, silently, and the nvim/vim diff upgrade is unreachable.

`use-file-open-actions.ts:55` (`["sh", "-c", …]` to the PTY host) was **unreachable** on Windows because the gate returned `null` first — a latent trap: fixing the probe alone converts a working fallback into an editor tab that spawns `sh` and dies. Hence one change, not two.

## The fix

- `src/lib/updater-shell.ts` → `src/lib/posix-shell.ts`, generalized (`posixShell()` / `missingPosixShellHint()`) rather than copied a third time. #956's two call sites move over unchanged in behaviour.
- The probe, `rove config`'s spawn, and the editor-tab argv all use it.
- **Path form**: strings interpolated into these commands now go through `toPosixPath()`. `C:\a\b` inside a Git Bash command line is escape soup; `/c/a/b` is understood by Git's own tools *and* converted back to a native path by the MSYS runtime when the target is a Windows binary. This covers `buildEditorCommand`, `buildNvimDiffCommand`, the auto-detect branch, and the diff-upgrade equality test (which would otherwise never match on Windows and silently drop diff mode).
- `rove config`'s failure message now adds the Git-for-Windows hint, so the advice is actionable.

No `where.exe` branch: probing through the shell we launch through is what makes `command -v` correct everywhere — Git Bash's PATH carries both the Windows entries and its own `/usr/bin`. Confirmed below (`notepad` resolves through it).

## Verified on a real Windows install

```
platform      : win32
posixShell()  : C:\Program Files\Git\bin\bash.exe
binaryAvailable(vim)             : true      <- Git Bash's own
binaryAvailable(nvim)            : false     <- genuinely not installed
binaryAvailable(notepad)         : true      <- native Windows binary, same probe
binaryAvailable(nonexistent-xyz) : false
resolved cmd  : {"bin":"cat","command":"cat '/c/Users/…/state.json'"}
--- now the real `rove config` ---
{}
EXITCODE=0
```

`EDITOR=cat` makes the launch observable: `rove config` really spawned the editor through Git Bash, on the right file, and propagated its exit code. Auto-detect (no `$EDITOR`) resolves to `vim '/c/Users/…/state.json'`.

That empirical run caught a bug in the first draft of this change: the auto-detect branch composes its own command string and was still emitting a backslash path. Fixed, then re-verified.

Not verified: interactive rendering of Git Bash's MSYS `vim` when launched from a Windows console. GUI editors (`notepad`, `code`) and non-interactive commands are confirmed working.

## Test

`bun run test:fast` 500 files / 4948 passed; typecheck + biome clean. New win32 path-form tests thread an explicit `platform` so they measure something on POSIX CI too, and `test/lib/posix-shell.test.ts` carries over #956's coverage under the new name.


ZHallen122:
CI green, 10/10, after two fixes worth recording:

**The `windows` job caught a Windows-only test break that `test:fast` cannot see.** `editor-launch-resolve.test.ts` stubs `Bun.spawn` and matched `cmd[0] === "sh"` — so on the one platform this whole change exists for, every probe fell through to the catch-all and three tests failed. The stub now matches `posixShell()`, plus an explicit assertion that the probe spawns it. Swept the rest of `test/` for the same pattern: the other `"sh"` literals are input data (plugin manifests, exec-host argv), not shell-choice assertions.

**One unrelated flake:** `terminal-pty-scrollback-cache.test.ts > converts ~live-grid rows per refresh` failed with `expected 210 to be less than 60` — the cache reading cold, i.e. its `settle()` not having landed on a slow runner. It passed in this PR's previous run and touches no code in this change (no xterm/scrollback here). Green on re-run.

PR #1022, opened 2026-09-17
Revert "feat: IBM Bob as a built-in engine"

Sma1lboy:
Reverts Sma1lboy/rove#1021


Sma1lboy:
History rewritten instead; main reset to 0.9.202.

PR #1023, opened 2026-09-17
revert: remove IBM Bob as a built-in engine

Sma1lboy:
Reverts #1021 (merged without review). Bob can return through a reviewed PR.

- Straight `git revert` of 2f52a8fec, no conflicts.
- Skill version goes **forward** to 47, not back to 45: `skill-install.ts` only marks an install stale when its marker is lower than the bundled one, so a 46 install would keep the Bob-era skill.
- Changeset: patch.

Checked locally: vitest `test/architecture test/engine test/types` (1034 passed), typecheck, root lint.


Sma1lboy:
History rewritten instead; main reset to 0.9.202.

PR #1048, opened 2026-09-19
fix: wait for a task's processes to exit before removing its worktree, so Windows can delete the directory

ZHallen122:
## What

Deleting a task on Windows left an empty `~/.rove/worktrees/<repo>/<name>` directory behind, every time, with `Permission denied` in the deletion audit line. On one machine: 8 of them in a week, plus 5 more from the same cause with other git errors.

The deletion runner (and `worktree.remove`, and land) already tore the session down before `git worktree remove`. Two things made that teardown not enough:

1. **It didn't wait.** `pty.kill` answered `accepted` as soon as the request was taken, so the removal ran while the child was still exiting — and a process whose cwd is inside the worktree makes the directory undeletable on Windows, so git deregistered the worktree and the directory stayed. `pty.kill` now takes `wait: true`, on which the PTY server holds the reply until the child's exit (bounded by the existing SIGTERM → SIGKILL grace), and the daemon-side teardown adapter passes it. An older host ignores the flag and answers as before.
2. **It reached the shell alone.** Ending a session on Windows was node-pty's `kill()` — `TerminateProcess` on the shell. ConPTY has no process groups, so the engine and everything it spawned survived with their cwd in the worktree. The node-pty driver now supplies `endTree` (`taskkill /T /F /PID`), and `terminatePtyChild` runs it *before* releasing the pseudo console. Keyed on the driver rather than `process.platform`: a fake child in a test carries a made-up pid, and a platform-keyed `taskkill /F` would reach whatever real process holds it.

`pty-server.ts` crossed the 500-line cap with the `wait` branch, so its verb dispatch moved to `pty-server-verbs.ts` (wire/lifecycle in one file, what each verb does in the other).

The first commit also added a teardown inside `TaskDeletionCoordinator.finish()`; the second removes it again — the runner already does that, so it only ran the teardown twice.

## Tests

- `pty-termination.test.ts`: tree-first ordering, a driver without `endTree` untouched, a failed tree kill still settles.
- `node-pty-driver.test.ts`: the driver exposes `endTree` aimed at the child's pid.
- `pty-kill-wait.test.ts` (new, runs on a named pipe on Windows): `wait: true` replies after the exit with `ended: true`; the plain verb still replies before it.
- `hosted-session` / `daemon-session-adapter`: the wait flag reaches the wire from the daemon-side teardown.

Verified locally on Windows 11: typecheck, biome, `test:fast` (the three remaining failures — `layout-migration` symlinks, `history-default-deps` mtime — fail identically on untouched `main`), and the socket suites that can listen on this platform. First-commit CI was green on every job including `windows`.

Docs: `docs/WORKTREES.md` gains a paragraph under "When git removes the worktree but not its directory" saying what a `Permission denied` residue means now.


wisp-agent-ai (approved):
Reviewed: CI green, local lint/typecheck + changed test files pass on head 5c87450.

PR #1051, opened 2026-09-19
feat: copilot and droid report their live session through hooks

Sma1lboy:
Copilot carried a `NoopHookAdapter` and droid carried no adapter at all, so both were screen-only: Rove could badge what they were doing but never learn WHICH engine session was live in a worktree. Each now installs one `SessionStart` observer, and their screen manifests keep owning working/blocked/idle exactly as before — the hook adds identity on top, the same split cursor landed on in #1032.

| engine | file | event → verb |
|---|---|---|
| copilot | `~/.copilot/settings.json` (`$COPILOT_HOME` honoured) | `SessionStart` → `session-start` |
| droid | `~/.factory/settings.json` (no env override) | `SessionStart` → `session-start` |

One event each, deliberately: neither engine is in refs/herdr's `full_lifecycle_hook_authority()` set, and herdr itself retreated from the nine lifecycle events it once installed on each to this one (its `*_REMOVED_LIFECYCLE_HOOK_EVENTS` lists are what it now deletes on sight).

## Two of the four engines in the brief are not in this PR

- **copilot did not fit `JsonHookAdapter`.** Its entries are FLAT — `{ "version": 1, "hooks": { "<event>": [ { "type": "command", "command": … } ] } }`, per [GitHub's hooks reference](https://docs.github.com/en/copilot/reference/hooks-reference) — where Claude, Codex and droid nest a `hooks` array inside each group. Rather than force it into the base or write a second copy of cursor's merge, the flat merge moved out of `cursor-local/hook-adapter.ts` into `engine/flat-hooks.ts` and the two engines share it. Cursor's public API and its test file are unchanged.
- **qodercli and devin are not Rove engines.** There is no entry for either in `builtin-engines.ts` or `CONTRIB_ENGINES`, so there is nowhere to hang an adapter — `activityHookAdapters()` walks registry entries. Adding them means a display name, a launch command and a translated screen manifest each, which is the "recognize X" shape of #1045 and outside this PR's layer-A fence.

`JsonHookAdapter` itself is untouched.

## Event-name note

Copilot accepts camelCase `sessionStart` and PascalCase `SessionStart` (the latter "for VS Code compatibility"). This installs PascalCase: the camelCase spelling has a standing not-firing report (github/copilot-cli#1730), and herdr installs PascalCase and deletes camelCase — the same conclusion from the field.

## Verification

- `bun run typecheck`, repo-root `bun run lint`, `bun run test:fast` (547 files / 5431 tests) all green.
- **Zero impact without the CLIs.** `rove api engine-list` under an isolated `HOME` is byte-identical to `origin/main`, and the run creates no `.copilot` and no `.factory` — both installs return early on a missing config dir.
- **Idempotent + non-destructive**, per engine: a second install is byte-identical on disk; a third party's entry (herdr's, including the `bash`-keyed form copilot writes), other events, and other top-level keys all survive; removal takes only Rove's entry. A dev-checkout install is REPLACED by a released one rather than stacked, because ownership is `isRoveHook`, not string equality.
- **Each adapter judges its own file** (#1032): copilot's flat validator accepts copilot's own valid document, which the Claude/Codex validator would call broken, and both adapters' `hookConfigRefusal` returns the same verdict the install abandons on.
- **Payload-driven round trip** (`test/cli/hook-install-roundtrip.test.ts`): install, read the command string back out of the file the engine would read, split it to argv, and feed that argv plus a hook payload to the real `runHookSubcommand`. The daemon receives `engine.reportEvent` with the payload's cwd, `session-start`, the right `--engine` tag and the session id. **Neither CLI is installed on the machine this was written on**, so the payloads are SHAPES — field names from GitHub's reference and from herdr's own copilot/droid hook scripts (`session_id`, with `sessionId` as copilot's fallback) — not captures. Same honesty caveat cursor's PR carried.
- **Mutation-checked**: dropping the ownership filter in the flat merge reddens 7 tests across copilot and cursor (idempotence, dev→released replacement, scoped removal); dropping droid's missing-dir guard reddens the "writes nothing at all" test.


Sma1lboy:
CI is green on the rerun. Attempt 1's render-track failure was `test/render/new-task-repo-name-field.test.tsx` → "typing an ambiguous name refuses to guess, and says so" — unrelated to this PR (new-task dialog synthetic input; `BUILTIN_VENDORS` and the dialog's engine row are untouched here). It passes 7/7 locally, and the failure frame showed the dialog rendered without the warning line yet, immediately followed by React "not wrapped in act(...)" warnings — the same shape as the render track's other synthetic-input flakes. Recording it here because `gh run rerun` hides attempt 1.

